PR #92, opened 2014-08-06
Added Timestamp datatype

skavanagh:
Uses current date if exist, but empty/null or if required in schema.


jewzaam:
Note, this will likely be closed as we complete work on a new generated field constraint (https://github.com/lightblue-platform/lightblue-core/issues/204)

bserdar:
No longer needed after generators.

PR #276, opened 2015-01-09
Fixes #19 - Json schemafor metadata

### LINKED ISSUE #19: RFE: Get JSON Schema for Metadata
We use json schemas internally, would be nice to provide the schema on demand to clients.  Kind of like a wsdl.  Maybe:

```
GET /metadata/<entity>/<version>?schema
```

luan-cestari:
Fixes #19


metadata/src/main/java/com/redhat/lightblue/metadata/FieldConstraint.java
```diff
@@ -31,4 +31,10 @@
      * Determines if the constraint is valid for the given field type
      */
     boolean isValidForFieldType(Type fieldType);
+
+
+    /**
+     * Returns the description of the FieldConstraint (for documentation propose)
+     */
+    String describeConstraint();
```
jewzaam: This name was confusing when I first saw it.  Maybe if it were `getDescription()` it would be clearer?

luan-cestari: I will change it. I used 'to describe' (the verb) intended to say that could be dynamic generated (using interpolation of different things to compose a meaning description) .

jewzaam:
Have only gotten part way through reviewing this.  Hopefully I can continue this afternoon but if not someone else is welcome to take it.

test/src/test/java/com/redhat/lightblue/test/metadata/FakeMetadataTest.java
```diff
@@ -36,6 +49,66 @@ public void testEntityInfo_VersionDoesExist(){
...
...e\":\"string\"},\"phoneNumber\":{\"type\":\"string\",\"matches\":\"^\\\\d{3}\\\\-\\\\d{4}\\\\ \\\\d{4}$\"},\"email\":{\"type\":\"string\"},\"company\":{\"type\":\"string\"},\"firstName\":{\"type\":\"string\"},\"requid\":{\"type\":\"uid\",\"required\":\"Field required constraint\"},\"emailConfirmed\":{\"type\":\"boolean\"}},\"updatedDate\":{\"type\":\"date\"},\"password\":{\"type\":\"object\",\"hashed\":{\"type\":\"string\"},\"salt\":{\"type\":\"string\"}},\"contactPermissions\":{\"type\":\"object\",\"allowMailContact\":{\"type\":\"boolean\"},\"allowThirdPartyContact\":{\"type\":\"boolean\"},\"allowPhoneContact\":{\"type\":\"boolean\"},\"allowFaxContact\":{\"type\":\"boolean\"},\"allowEmailContact\":{\"type\":\"boolean\"}},\"_id\":{\"type\":\"string\",\"identity\":\"Field is part of identity constraint\"},\"active\":{\"type\":\"boolean\"},\"login\":{\"type\":\"string\",\"maxLength\":\"64\",\"minLength\":\"1\",\"required\":\"Field required constraint\"},\"objectType\":{\"type\":\"string\"},\"requid\":{\"type\":\"uid\",\"description\":\"test\",\"required\":\"Field required constraint\"},\"createdDate\":{\"type\":\"date\"},\"required\":[\"login\",\"personalInfo.requid\",\"requid\"]}}";
```
jewzaam: Would be easier to manage in a test resource and loaded from disk.  Using skyscreamer we can easily verify without worrying about formatting, meaning the file can be pretty printed.

test/src/test/java/com/redhat/lightblue/test/metadata/FakeMetadataTest.java
```diff
@@ -36,6 +49,66 @@ public void testEntityInfo_VersionDoesExist(){
...
+        JSONAssert.assertEquals(expected, actual, false);
```
jewzaam: Should additionally verify that `actual` is a valid json schema.  We have a util that will [verify from a file resoruce](https://github.com/lightblue-platform/lightblue-core/blob/master/util/src/main/java/com/redhat/lightblue/util/JsonUtils.java#L122-L139).  Could update that class to have an api that supports passing in a json schema string for validation also.

jewzaam:
Moving to in progress, please comment when updates are ready for review and I'll work to look as soon as I can.

jewzaam:
@luan-cestari this needs updated still

luan-cestari:
updated the PR. @jewzaam

jewzaam:
Just noticed `FakeMetadataTest` is in the test module.  The unit tests for these changes should be in the metadata module.  This test is to test the `FakeMetadata` class.

Generated schema is taking constraints on metadata fields and converting them to properties on the simple objects.  For example, this:

```
      "iduid": {
        "type": "uid",
        "constraints": {
          "identity": true
        }
      },
```
becomes
```
        "iduid": {
            "type": "uid",
            "identity": "Field is part of identity constraint"
        },
```

It shouldn't do this.  Identity isn't part of the json schema spec.  I expected the json schema validation to catch this, but I guess it ignores additional properties.  Identity should be treated the same as required.  See comments below.

Also, required constraints should be reflected in a `required` field in the parent of the property in the json schema.  They shouldn't be collected and reported at the root level property set.  It's done today by using the lightblue full path, a period delimited string.

The "uid" type is converted directly to the json schema.  There is no "uid" type in the spec.  It should become a strong.

The "match" constraint is converted directly.  It should be instead converted to a "pattern" attribute on the object representing the field.

I think min and max constraints are converted correctly.

I do not see a unit test including enum.

Moving back to in progress.

luan-cestari:
Looking. The first change (about move to different module), I did but after I tested it I noticed that it can't be that way with those files, as the FakeMetadataTest class uses some classes from test module and test module have a dependency to the metadata module, so if I move the FakeMetadataTest class to metadata I would need the test module as dependency which would make a cyclic dependency issue. 

About the other task to do, I'm working on it. I updated the json schema to have the validation necessary. Just one small question (to check if I got what you mean) one the snippets that you gave as example, so the first one* would be the output expected to be generated, right?

By 'first one' I mean:
```
      "iduid": {
        "type": "uid",
        "constraints": {
          "identity": true
        }
      },
```

jewzaam:
If `iduid` was the only field in the metadata and lightblue `uid` type mapped to a json schema type of `string` without any regex:

```json
    "properties": {
        "iduid": {
            "type": "string"
        }
    },
    "required": [
        "iduid"
    ]
```

luan-cestari:
Status summary:
- [x] metadata/test module change (result: not moved dua cyclic dependency)
- [x] Generated schema is taking constraints on metadata fields and converting them to properties (result: last commits fix that)
- [x] json schema validation to catch problems (result: create a simple schema. Do you want to use the 'schema.json' file from the metadata instead? https://github.com/lightblue-platform/lightblue-core/blob/cc6f4ae09bdfadb9372a7ea7e2e4476ad8accca0/metadata/src/main/resources/json-schema/metadata/schema.json )
- [x] required constraints should be reflected in a required (result: last commits fix that and also for identity)
- [x] The "uid" type should become a string. (result: last commits fix that )
- [x] "match" constraint be converted to a "pattern" attribute on the object representing the field.(result: last commits fix that )
- [x] min and max constraints are converted correctly.
- [ ] unit test including enum.
- [ ] Other things?

luan-cestari:
@jewzaam I update the PR (also put a summary to help to track what have done and what we need to do) but I will focus on RDBMS integration test now, okay?

jewzaam:
@luan-cestari thank you, yes please focus on rdbms.  This can be dropped from Ana if necessary.

jewzaam:
@luan-cestari given shift of focus and your work on non-rdbms issues could you consider closing out this issue?

luan-cestari:
@jewzaam I just updated the PR to also cover the enum case. It was working fine, I just made some changes to also output the enum definition on the json as it is defined on teh entityinfo section which would not be included . The output would be something like https://gist.github.com/luan-cestari/a0ee0b58e9a08467d9a2 . Ready for review

luan-cestari:
By the way , com.redhat.lightblue.metadata.EnumValue seems to always have empty/null description, is this fiend required at all? Could it be exemplified in the documentation http://docs.lightblue.io/language_specification/metadata.html ?

jewzaam:
Created https://github.com/lightblue-platform/lightblue-docs-specs/issues/7 for the missing documentation.  Regarding it being empty, are you supplying descriptions for enum values?  If not, yes it should be empty.

luan-cestari:
It only accept and array of Strings. It seems that the description is only used when it is bind with a field called "annotatedValues" instead "values". I will try to use both cases and make an exception on the schema generation for that each case, ok?

jewzaam:
Let's discuss enum structure on https://github.com/lightblue-platform/lightblue-docs-specs/issues/7 since it needs documented.

luan-cestari:
Alright, I will wait the resolution of that discussion. Let me know the
changes that it is pending to be made on this PR.

luan-cestari:
I will make little updates to create the Rest service for this feature

metadata/src/main/java/com/redhat/lightblue/metadata/FieldConstraint.java
```diff
@@ -31,4 +31,10 @@
      * Determines if the constraint is valid for the given field type
      */
     boolean isValidForFieldType(Type fieldType);
+
+
+    /**
+     * Returns the description of the FieldConstraint (for documentation propose)
+     */
+    String getDescription();
```
jewzaam: Why do we need documentation on the constraint?  There is nothing in metadata to describe why we set a constraint other than what might be saved in the field's description.  But that's part of the field, not the constraint.

luan-cestari: The documentation you  mean the getDescription method right? This method is just a way to use polymorphism so I can get the specific description of each constraint without casting or checking instanceof. The description is used to generate a user-friendly message in the json schema

jewzaam: Why does the constraint need a description?  Each property in the json schema needs a description, yes, but those are given by the field in lightblue metadata.  Constraints don't have a description in json schema.

luan-cestari: Maybe I chose the wrong word for the method name, it basically helps to generate an user text for a constraint and it can vary a lot. After the iterations we had for the issue that this PR aism to solve, the regex constraint (which uses 'pattern') is the only one using it right now and on this case it use to describe the pattern used for that constraint (which can vary for each request)

jewzaam: It might need to be specific to each constraint.  Numeric value constraints (minimum, maximum, minLength, etc) will be a number in json (no quotes) vs a pattern which will be a string (with quotes).  I haven't reviewed how this is used yet to see.  Interestingly ArrayElementIdConstraint returns a bunch of text, so I assume given your comment this is broken or it isn't used in a generic way.

luan-cestari: It was used, by I just quickly  reviewed it before we talk about it, but I think it is just the regex case now, but it was used for other cases as well (when I had a different concept of the schema json that we would generate), so I can refactor it to something different just for regex case now if you think it would be better

PR #279, opened 2015-01-13
Fix #208: check field access for query fields

### LINKED ISSUE #208: If a caller does not have access to a field, that field cannot be used in a query.

bserdar:


crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -418,4 +423,28 @@ private void updatePredefinedFields(OperationContext ctx,CRUDController controll
...
+            LOGGER.debug("Checking access for query fields {}",fields);
+            for(FieldInfo field:fields) {
+                LOGGER.debug("Access checking field {}",field.getAbsFieldName());
+                if(eval.hasAccess(field.getAbsFieldName(),FieldAccessRoleEvaluator.Operation.find)) {
+                    LOGGER.debug("Field {} is readable",field.getAbsFieldName());
+                } else {
+                    LOGGER.debug("Field {} is not readable",field.getAbsFieldName());
+                    ctx.addError(Error.get(CrudConstants.ERR_NO_ACCESS, field.getAbsFieldName().toString()));
+                    ctx.setStatus(OperationStatus.ERROR);
+                    return false;
```
jewzaam: Since we have all the fields, why not continue processing to find out if there are additional things the caller is not authorized to access, add each as an error?

bserdar: ok, will add that.

bserdar: Ready for review.

PR #296, opened 2015-01-20
Travis update

luan-cestari:
The project is now following the same configuration as the https://github.com/lightblue-platform/lightblue-rest (but with a different encrypted key/pass to deploy to remote repository).


jewzaam:
Should merge https://github.com/lightblue-platform/lightblue-core/pull/294 first?

jewzaam:
@luan-cestari something with the coveralls:cobertura goal is failing now.  I cannot reproduce locally because it's only when the coveralls goal runs and that requires publishing metrics.  Seems it's trying to get coverage for the root pom but there is nothing to report.  I think we saw this a while ago but it was with jacoco.  The differences from what it used to be that I see after merging master are:
1. added -B to mvn execution
2. added verify goal to mvn execution
3. added properties to mvn execution

I don't think 3 would be a problem.  It is probably not necessary for this repo.  That leaves 1 and 2.  Is it related to the PR issues with ldap repo?

PR #297, opened 2015-01-22
Fixes #283: Array field comparison

### LINKED ISSUE #283: enhance nary_comparison to support referencing a field instead of a static list of values
A problem for many-to-many associations. Cannot do today without having a separate entity that makes it a 1:m and m:1 (one entity too much if you ask me).

bserdar:
Adds support for 
  - Nary field relational expression
  - Arrays for field comparison expression


jewzaam:
@bserdar is this to fix https://github.com/lightblue-platform/lightblue-core/issues/283 ?

jewzaam:
Nevermind, I see it in the title now.  Updating to link it in waffle.

bserdar:
Yes

luan-cestari:
It seems fine, we could merge but I would suggest some enhances (some may be not part of this PR or event eh issue itself but maybe it simple enough to be done together, otherwise we can create new issues for them) .

- [] Use Enums for constants
- [] Log the exceptions/errors (also maybe increase the log debug on some places)
- [] I noticed that the test coverage decreased. I would suggest we could increase the coverage in some classes like metadata/src/main/java/query-api/src/main/java/com/redhat/lightblue/query/RelativeRewriteIterator.java , metadata/src/main/java/crud/src/main/java/com/redhat/lightblue/assoc/ResolvedFieldBinding.java , metadata/src/main/java/crud/src/main/java/com/redhat/lightblue/assoc/qrew/rules/EliminateNOT.java , metadata/src/main/java/crud/src/main/java/com/redhat/lightblue/eval/FieldAccessRoleEvaluator.java , metadata/src/main/java/crud/src/main/java/com/redhat/lightblue/eval/ArrayAddExpressionEvaluator.java
 - [] Rename the resources files such as https://github.com/bserdar/lightblue-core/tree/array-field-comparison/crud/src/test/resources to follow some pattern (probably another issue as we would come up with a nice pattern but I think the files should be easy to be recognisable) (for comparison, the files in https://github.com/bserdar/lightblue-core/tree/array-field-comparison/crud/src/test/resources/composite )

bserdar:
Added new tests for RelativeRewriteIterator. The remaining classes you mentioned are not related to this PR.

bserdar:
Use Enums for constants : Which constants?

bserdar:
Log the exceptions/errors (also maybe increase the log debug on some places): Error.get logs the error when called with an exception, so extra logging is not necessary there.

bserdar:
Rename the resources files...: Created #309

luan-cestari:
About Enum: for example, the class crud/src/main/java/com/redhat/lightblue/assoc/AssocConstants.java could be an enum .

About the error messages &log, you are right, I just thought that some cases we could use log debug to put much more context about what is happening (especially information that we don't want to put on the error message we are going to send to the user), but for me it seems fine the way it is.

Could you update the PR due it can't be merged right now, please?

bserdar:
AssocConstants.java: Those are error codes, why would those be an enum?

PR is updated.

luan-cestari:
About the enums, no problem, it would just be almost the same as it is now.

For me, the way it is now good. I will merge it

PR #301, opened 2015-01-26
Fix travis deploy

luan-cestari:
Fixed the onSuccess.sh  and add the oracle jvm 7 to be tested. Note This change (which can only be tested after merged on master branch) should be replicated to other repositories (as the problem doesn't make changes to be deployed, in other words, the remote maven repositories are outdated with the latest changes)


jewzaam:
Why did we add oraclejdk7 ?

PR #303, opened 2015-01-26
Entity metadata validation only checks top level fields. When this is fi...

bserdar:
...xed, validation fails during CompositeMetadata construction, so removed validation during construction as well.


jewzaam:
@bserdar could the issue you saw with CompositeMetadata construction be reproduced?

bserdar:
Yes call validation in entitymetadata ctor and lots of tests will fail.

bserdar:
Added unit tests

jewzaam:
@bserdar was the problem then second level enum validation, and it just manifested itself in composite metadata?

bserdar:
First problem was the validate function. It got the root-level fields and
iterated through them, but never validated anything deeper.

Second problem was the fact that validate() was called in the ctor. During
composite metadata construction, EntityMetadata objects are constructed,
and then they're tied together. During construction, the field tree is not
valid, and when validate() tried to iterate through broken links, it threw
element not found exceptions.

PR #306, opened 2015-01-28
Fix #305: empty object fields are not iterable

### LINKED ISSUE #305: AbstractTreeCursor#firstChild threw exception instead of returning false
I crreated a unit test for an ObjectField that didn't actually have any children fields. Instead of #firstChild returning a false, it threw the below exception. This seems to be because ObjectField#hasChildren always returns true. Then the #firstChild method blindly attempts to grab the next node, which is what causes this exception. Seems like KeyValueCursor#hasNext might need to be consulted prior to calling #next, if #hasNext is false, then return false.

```
java.util.NoSuchElementException
	at java.util.ArrayList$Itr.next(ArrayList.java:834)
	at com.redhat.lightblue.metadata.FieldCursor$Adapter.next(FieldCursor.java:45)
	at com.redhat.lightblue.util.AbstractTreeCursor.firstChild(AbstractTreeCursor.java:189)
	at com.redhat.lightblue.crud.ldap.translator.ResultTranslator.toJson(ResultTranslator.java:170)
	at ...
```

bserdar:


dcrissman:
Fixes #305

PR #308, opened 2015-02-09
Push  skeleton of query iterator to a base class so extending clas...

bserdar:
...s can deal with different return types


luan-cestari:
I saw the code, I think that the ' else ' statement should be like the previous last ' else if ' ( " else if (q instanceof AllMatchExpression" ) with a new else statement next to it to deal with the case we forgot to implement a new kind of expression (so it would be help in the development/enhancement over our own api , it wont be reachable from the client interface)

bserdar:
Fixed, I think it is better now.

PR #310, opened 2015-02-09
Fixes #293: statsd metric observer added, updated servo graphite setup impl to use it

### LINKED ISSUE #293: Create our own metrics publisher for Graphite
Since we are having trouble with using Servo to publish Hystrix metrics to Graphite.  Essentially this just needs to push text to a socket in the following format

prefix.metricName value timestamp\n

jewzaam:
Puppet changes required to use, tracking in https://github.com/lightblue-platform/lightblue-puppet/issues/93

### DIFF
--- a/util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
+++ b/util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
@@ -110,12 +110,22 @@ public final class ServoGraphiteSetup {
             }
         }
 
-        if (port == null) {
-            port = "2004"; //default graphite port
-            LOGGER.debug("Using default port: " + port);
+        int iport = -1;
+        if (port != null && !port.isEmpty()) {
+            try {
+                iport = Integer.valueOf(port);
+            } catch (NumberFormatException e) {
+                iport = -1;
+                LOGGER.warn("Configured port is not an integer.  Falling back to default");
+            }
         }
 
-        String addr = host + ":" + port;
+        if (iport < 0) {
+            iport = 2004; //default graphite port
+            LOGGER.debug("Using default port: " + iport);
+        }
+
+        String addr = host + ":" + iport;
 
         LOGGER.debug("GraphiteMetricObserver prefix: " + prefix);
         LOGGER.debug("GraphiteMetricObserver address: " + addr);
@@ -146,11 +156,13 @@ public final class ServoGraphiteSetup {
         LOGGER.debug("STATSD_PORT environment variable is: " + port);
 
         int iport = -1;
-
-        try {
-            iport = Integer.valueOf(port);
-        } catch (NumberFormatException e) {
-            iport = -1;
+        if (port != null && !port.isEmpty()) {
+            try {
+                iport = Integer.valueOf(port);
+            } catch (NumberFormatException e) {
+                iport = -1;
+                LOGGER.warn("Configured port is not an integer.  Falling back to default");
+            }
         }
 
         if (iport < 0) {
--- a/util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
+++ b/util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
@@ -63,10 +63,7 @@ public final class ServoGraphiteSetup {
     }
 
     public static void initialize() {
-        // Without GRAPHITE_HOSTNAME variable, graphite will not start
-        String env = System.getenv("GRAPHITE_HOSTNAME");
-        LOGGER.debug("GRAPHITE_HOSTNAME environment variable in initialize() is " + env);
-        if (!initialized && env != null && !env.trim().isEmpty()) {
+        if (!initialized) {
             doInitialize();
         }
     }
--- a/util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
+++ b/util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
@@ -40,6 +40,13 @@ import java.util.concurrent.TimeUnit;
 public final class ServoGraphiteSetup {
     private static final Logger LOGGER = LoggerFactory.getLogger(ServoGraphiteSetup.class);
 
+    public static final String ENV_GRAPHITE_PREFIX = "GRAPHITE_PREFIX";
+    public static final String ENV_GRAPHITE_HOSTNAME = "GRAPHITE_HOSTNAME";
+    public static final String ENV_GRAPHITE_PORT = "GRAPHITE_PORT";
+    public static final String ENV_STATSD_PREFIX = "STATSD_PREFIX";
+    public static final String ENV_STATSD_HOSTNAME = "STATSD_HOSTNAME";
+    public static final String ENV_STATSD_PORT = "STATSD_PORT";
+
     private static boolean initialized = false;
 
     private ServoGraphiteSetup() {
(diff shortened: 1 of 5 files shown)


util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
```diff
@@ -68,19 +71,28 @@ public static void initialize() {
...
+        String port = System.getenv("GRAPHITE_PORT");
+
+        // verify at least hostname is set, else cannot configure this observer
+        if (null == host || host.trim().isEmpty()) {
+            LOGGER.info("GraphiteMetricObserver not configured, missing environment variable: GRAPHITE_HOSTNAME");
             return;
         }
 
-        // register hystrix servo metrics publisher
-        HystrixPlugins.getInstance().registerMetricsPublisher(HystrixServoMetricsPublisher.getInstance());
```
luan-cestari: I think it would be good to keep this line if it can help us in case we made any change that would use Hystrix command prior than this execution (as mentioned in the netflix issue)

util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
```diff
@@ -98,31 +110,89 @@ private static synchronized void doInitialize() {
             }
         }
 
-        String host = System.getenv("GRAPHITE_HOSTNAME");
-        LOGGER.debug("GRAPHITE_HOSTNAME environment variable in doInitialize() is " + host);
-        String port = System.getenv("GRAPHITE_PORT");
-        LOGGER.debug("GRAPHITE_PORT environment variable in doInitialize() is " + port);
         if (port == null) {
```
luan-cestari: I would suggest that it follows the same verification as the statd block, using Integer.valueOf and NumberFormatException

util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
```diff
@@ -98,31 +110,89 @@ private static synchronized void doInitialize() {
...
+        LOGGER.debug("STATSD_PREFIX environment variable is: " + prefix);
+        LOGGER.debug("STATSD_HOSTNAME environment variable is: " + host);
+        LOGGER.debug("STATSD_PORT environment variable is: " + port);
+
+        int iport = -1;
+
+        try {
+            iport = Integer.valueOf(port);
+        } catch (NumberFormatException e) {
+            iport = -1;
```
luan-cestari: Better log a message here as it might do something unexpected (for instance, if I initialized the environment variable  with a single letter in the end of it by mistake I could lose some time to figure out that the statd is working in a different port )

util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
```diff
@@ -98,31 +110,89 @@ private static synchronized void doInitialize() {
...
+        }
+
+        if (iport < 0) {
+            iport = 8125; //default statsd port
+            LOGGER.debug("Using default port: " + port);
```
luan-cestari: If the change mentioned previous comment about the port be implemented, I think this log will not be so meaningful

util/src/main/java/com/redhat/lightblue/util/ServoGraphiteSetup.java
luan-cestari: Noticed that it moved to the doInitialize() method, so you may disconsider this comment or move the line back to here if you think it would be better.

util/src/main/java/com/redhat/lightblue/util/statsd/StatsdMetricObserver.java
```diff
@@ -0,0 +1,72 @@
...
+        this.namingConvention = namingConvention;
+    }
+
+    @Override
+    public void updateImpl(List<Metric> metrics) {
+        // The statsd client doesn't do any checks on the underlying socket's state
+        // and the socket connects only once, so we cannot trust the socket to stay
+        // open over a period of time.  If this is changed/fixed we could reuse the
+        // client but until then it cannot be safely reused.
+        StatsDClient statsd = new NonBlockingStatsDClient(prefix, host, port, new ErrorHandler());
```
luan-cestari: Could ErrorHandler be a singleton?

util/src/main/java/com/redhat/lightblue/util/statsd/StatsdMetricObserver.java
```diff
@@ -0,0 +1,72 @@
...
+        LOGGER.debug("sending data");
+
+        for (Metric metric : metrics) {
+            String aspect = namingConvention.getName(metric);
+
+            // TODO figure out how to determine if this is a guage vs counter
```
luan-cestari: Looking the source code of servo, I would say that the metrics that is gauge  (extending com/netflix/servo/monitor/Gauge.java ) would have a tag DataSourceType.GAUGE ( com/netflix/servo/annotations/DataSourceType.java) on it . You could also check for the DataSourceType.COUNTER tag as well

luan-cestari: We could file an issue about it to enhance their wiki as well

luan-cestari:
I commented some points of the code, if you could review them please @jewzaam thanks

jewzaam:
Updated per comments.
Note in https://github.com/lightblue-platform/lightblue-core/commit/8ec6e37e2a6deaf528050a78bafd1ab39134a051 I realized initialization wouldn't allow for just a statsd config so it's fixed.

I will add some unit tests.  If this gets merged before I get the unit tests in I'll put in a new PR for those tests.

luan-cestari:
I will wait for the units test but it seems very nice

jewzaam:
@luan-cestari struggling with how to do any real testing of the observer.  I would like to verify the logic for deciding counter vs gauge is sound, but it requires so much setup to get metrics collected I think I'll have to pass.

I will add at least some initialization tests for the setup class.

luan-cestari:
@jewzaam OK. I would suggest to use a classloader instance for each test as you are dealing with some static methods. Maybe that junit runner extension that I pushed for rest repository could help (but it can only handle one classloader for the entire junit class, so you would need to create some classes instead methods on the junit class to avoid them to use the same classloader  ).

jewzaam:
@luan-cestari tests added as much as I could.  One of the issues is our use of env variables, which are immutable in java.  So, the register methods were updated to take the prefix, host, port as args and therefore could be tested.

jewzaam:
@luan-cestari for the tests I created I don't see why a classloader is necessary.  The methods under test are static, but there are no changes made to state other than additional items added to the `observers` list that is an argument.  If I'm missing something please let me know.

luan-cestari:
About the issue on tests, maybe we could rely in one of the hacks mentioned on https://stackoverflow.com/questions/318239/how-do-i-set-environment-variables-from-java (I already tested than, they work).

Abut the classloader, sorry forgot to mention, I was thinking about the Hystrix issue, make that reproducible in the tests so we can be confident that other statement that check if it already started is working or not (I'm almost sure it is but I think it is a nice test). (The detail about the classloader here is that Hystrix must rely on something static which is bound to the classloader, so we initialize it  using an specific controlled classeloader so we can have different tests, like some of them the Hystrix command can be called earlier and other later than that method). Let me know if it is not very clear yet, so I can maybe give more details or even write a single test about it.

jewzaam:
Yes, classloader makes sense for the hystrix testing.  I will not be including those tests in this PR.

As for env, I didn't notice the second (non-answer) response earlier today.  Looking into it now it looks promising but the hacks for testing this are getting ugly.  Now I'm hitting against classloader issues because of the HystrixServoMetricsPublisher.  It gets initialized once and there's no way to reset it.  And we explicitly want it to fail if multiple attempts to register it happen, so no error handling can be done.

jewzaam:
@luan-cestari after further playing around with the classloader isolation I'm going to stop.  This isn't something I feel we'll get enough value out of right now.  If you're really keen on getting it done, please open an issue.  This PR is ready for final review (meaning I don't plan on adding more tests).

PR #328, opened 2015-02-20
create utility for creating instance of LightblueFactory for testing purpose

dcrissman:
This is an abstraction of a testing utility I created for LDAP in order to make it more generally available.

### DIFF
--- a/test/src/main/java/com/redhat/lightblue/test/AbstractCRUDTestController.java
+++ b/test/src/main/java/com/redhat/lightblue/test/AbstractCRUDTestController.java
@@ -35,11 +35,14 @@ import com.redhat.lightblue.metadata.EntityMetadata;
 import com.redhat.lightblue.metadata.Metadata;
 
 /**
- * Testing harness for junit tests that need to stand an in-memory instance of lightblue.<br>
+ * <p>Testing harness for junit tests that need to stand an in-memory instance of lightblue.
+ * The assumption is made that one {@link LightblueFactory} will be used per test suite.</p>
+ * <p>For initialization code that needs to be executed prior to the {@link #AbstractCRUDTestController()},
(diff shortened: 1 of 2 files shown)


dcrissman:
This is ready for review.

jewzaam:
The createLightblueFactory API also is creating metadata.  Lightblue factory creation exists in rest/common class RestConfiguration but doesn't allow for overriding datasource filename.  The methods for creating request and response seem like something that could be added to an abstract test class.  Didn't look to see if we have something already that would be appropriate.  Having a util class isn't wrong, but might be harder to find.

dcrissman:
I have no issue using an abstract class over a util.

test/src/main/java/com/redhat/lightblue/test/AbstractCRUDTestController.java
```diff
@@ -0,0 +1,122 @@
...
+     * @param metadataResourcePaths - path to any and all *-metadata.json you'd like {@link LightblueFactory} to know about
+     * @return an instance of {@link LightblueFactory}.
+     * @throws IOException
+     * @throws ClassNotFoundException
+     * @throws NoSuchMethodException
+     * @throws IllegalAccessException
+     * @throws InvocationTargetException
+     * @throws InstantiationException
+     */
+    protected static LightblueFactory createLightblueFactory(String datasourcesResourcePath, String... metadataResourcePaths)
```
jewzaam: Should this be executed by some @BeforeClass annotated method?  Could make the arguments abstract getter methods.  This way the factory is always created.

dcrissman:
@jewzaam Change made, ready for review.

luan-cestari:
Seems pretty nice. I will merge.

PR #330, opened 2015-02-20
add custom asserts to check lb Response for Errors or DataErrors

dcrissman:


test/src/main/java/com/redhat/lightblue/test/Assert.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Asserts no Errors are on a {@link Response} from lightblue.
+     * @param response - {@link Response} to check.
+     * @throws MultipleFailureException
+     */
+    public static void assertNoErrors(Response response) throws MultipleFailureException{
+        List<Throwable> errors = new ArrayList<Throwable>();
+        for(Error error : response.getErrors()){
+            Exception e = new Exception(error.getMessage(), error);
```
jewzaam: Error extends RuntimeException, this is unnecessary overhead.

test/src/main/java/com/redhat/lightblue/test/Assert.java
```diff
@@ -0,0 +1,70 @@
...
+            e.printStackTrace();
```
jewzaam: Two suggestions:
* if printing to standard out, pass a boolean argument to control if the stack trace is printed or not
* consider using a logger instead, which could be configured to write to a file or controlled with logging level.

PR #332, opened 2015-02-24
Adding webhook for buildtimetrend 

jewzaam:
https://buildtimetrend.herokuapp.com/#get_started

Just trying it out.  Don't merge.


jewzaam:
Might as well make this part of the build.  If someone objects we can discuss it here?

PR #338, opened 2015-03-06
Fixes #236 - Allow ANY path in $foreach

### LINKED ISSUE #236: Allow ANY path in $foreach
If updating an array inside an array you have to do nested $foreach statements.

luan-cestari:
Changes necessary for the enhancement https://github.com/lightblue-platform/lightblue-core/issues/236


bserdar:
Few things:
 - generateProcessedInfo doesn't need factory, pls remove from arg list
 - Pls rename 'ProcessedInfo' to something more meaningful, could be UpdateInfo
 - updateUsingProcessedInfo() 'ret' should not be an arg
 - Rename updateUsingProcessedInfo to update()

PR #341, opened 2015-03-11
Moved hystrix/servo/statsd/graphite config from util to new module: hystrix

jewzaam:
The idea here is to reuse this new module in other places without pulling along other dependencies.  Specifically, the lightblue-client for https://github.com/lightblue-platform/lightblue-client/issues/57 will be adding use of hystrix commands and needs this functionality.  I don't want to recreate it.  Another option if this is not desired is to split this out into a new repo so it's completely separate from lightblue-core.

Subsequent PR for lightblue-rest will be put in to use this new dependency.

### DIFF
--- a/hystrix/pom.xml
+++ b/hystrix/pom.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!--
+    Copyright 2013 Red Hat, Inc. and/or its affiliates.
+
+    This file is part of lightblue.
+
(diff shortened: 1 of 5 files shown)


jewzaam:
After this is merged and travis builds please look at https://github.com/lightblue-platform/lightblue-rest/pull/111 to force it to rebuild and see if it can be merged.

jewzaam:
Travis hasn't seen the merge of this PR.  And a few others.  Looks like going to the webhook config for the Travis CI integration and doing a "Test Service" will force a build though.

PR #346, opened 2015-03-16
Fix AbstractCRUDTestController

dcrissman:
Allows the AbstractCRUDTestController to be executed in a static or non-static way, ultimately depends on the test suites needs.


dcrissman:
Additional changes to #328

PR #352, opened 2015-03-23
Create CONTRIBUTING.md

luan-cestari:
Using GIthub feature (https://github.com/blog/1184-contributing-guidelines) that enable a notification before PRs and Issues be created which helps we add more value to those issues/PRs

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+# Guide lines for contribuition in our gitbook!
+* [http://dev.docs.lightblue.io/development_process/README.html](http://dev.docs.lightblue.io/development_process/README.html)
+* Questions & suggestions,s end on nabble ->  http://forum.lightblue.io/ OR  http://dev.forum.lightblue.io/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,3 @@
-# Guide lines for contribuition in our gitbook!
-* [http://dev.docs.lightblue.io/development_process/README.html](http://dev.docs.lightblue.io/development_process/README.html)
-* Questions & suggestions,s end on nabble ->  http://forum.lightblue.io/ OR  http://dev.forum.lightblue.io/
+# Guide lines for contribuition in our Gitbook!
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,3 @@
+# Guide lines for contribuition in our gitbook!
+* [http://dev.docs.lightblue.io/development_process/README.html](http://dev.docs.lightblue.io/development_process/README.html)
```
jewzaam: Some other text would be nice here instead of the URL.  Maybe "Development Process".

bserdar:
There's a few typos:
 - guidelines is one-word
- contribuition -> contribution
--> suggestions,s end on nabble ->

CONTRIBUTING.md
```diff
@@ -0,0 +1,3 @@
...
+* Questions & suggestions,s end on nabble ->  http://forum.lightblue.io/ OR  http://dev.forum.lightblue.io/
```
jewzaam: /s end/send/

And suggestions for contribution would be for the dev forum only.  Recommend making it a nice like like:

```
Questions & suggestions?  Share on our [dev forum](http://dev.forum.lightblue.io/)!
```

CONTRIBUTING.md
luan-cestari: Fixed

CONTRIBUTING.md
luan-cestari: Fixed.

Thank you for both comments. 

The only reason that I put the http://forum.lightblue.io/ there is because the user that was about to create an issue about something related to usage could also see this page (just some more details for clarification: many projects in GIthub people interact exclusively by GIthub issues, even about usage or asking for a feature (that could even already have), so this group of people could assume that lightblue follow this pattern as well but when they see the new issue page they will have that notification box which would lead to the page rendered from the content of this file, so some user related links might be helpful as well (maybe even more helpful if we ask to search in the documentation and nabble and then create a thread on nabble) (I think the documentation also cover this point, but just wonder the best way to also inform the lightblue's users about those best practices )  )

luan-cestari:
Thanks for the reviews =)

PR #354, opened 2015-03-25
Association nested array projection

bserdar:


paterczm:
Tested association in an array in an array in a real life scenario. Merging this in.

paterczm:
Closed it accidentally, pls ignore.

PR #366, opened 2015-04-02
wrap the stats connection in a try/catch

dcrissman:
Just ensuring the close


luan-cestari:
Asked to travis to build this PR again as the dependency that we used for this project changed and it can effect

luan-cestari:
It worked fine

PR #367, opened 2015-04-02
bump version of java-statsd-client

dcrissman:


luan-cestari:
Just a note about this PR: This is needed as not only the newer version might have bug fixes (and even new features) but mostly because we experienced some memory/thread/connection issues and this might help as it change a little bit the internal components around the connection.

PR #388, opened 2015-04-27
Fix #226: Add error context info to composite metadata construction

### LINKED ISSUE #226: Add Error push/pop to CompositeMetadata
There's no context added, which wasn't a big deal because there were no lightblue Error's thrown.  But I am about to introduce on in a PR so it's useful to have some context in this class (and maybe others) about what is going on.

bserdar:


paterczm:
@bserdar, can you provide a description? I don't know what this change is just by looking at it.

bserdar:
Added Error.push()/Error.pop to construction methods, so if it fails, the
error thrown will have the proper context information in it.,

PR #391, opened 2015-04-30
Fixes lightblue-platform/lightblue-audit-hook#38 : Hook problem - audit currently records all fields during an update, not just the modified fields

luan-cestari:
The issue https://github.com/lightblue-platform/lightblue-audit-hook/issues/38

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/crud/DocCtx.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/DocCtx.java
@@ -49,6 +49,7 @@ public class DocCtx extends JsonDoc {
     private final List<Error> errors = new ArrayList<>();
     private JsonDoc outputDoc = this;
     private JsonDoc originalDoc = null;
+    private JsonDoc originalOutputDoc = null;
     private CRUDOperation CRUDOperationPerformed;
     private final Map<String, Object> propertyMap = new HashMap<>();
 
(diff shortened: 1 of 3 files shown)


bserdar:
This change has too much untested ramifications. You are changing a lo of
conventions around DocCtx, without changing the comments around it.

Why do you need an originalOutputDoc? DocCtx keeps the old version of the
doc, and the new version. What is the originalOutputDoc?

luan-cestari:
@bserdar I agree about the ramifications, but most of them we have unit test and integration tests to assert that it is working. We can make more tests if you are very concerned about any problem, I would just like you to mention the most important tests that you would expect so we can move forward more quickly.

About the PR (like why the need of the originalOutputDoc field). What was actually happening in the DocCtx was it would have the old value and the PROJECTED new value. As removing the projection would cause much more changes, I chose to add a new field.

bserdar:
I understand the problem: we need :
 - The original document read from the db.  This would be null for insert.
 - The updated document. The unprojected version.
 - The projected document. 
The right way to do this is to change how outputDoc is used. Wherever we referred to the output doc, it should mean the projected doc. The new field should be named "updatedDoc" unless you have a better suggestion, and should contain the onprojected updated version. It should be set in update/save implementation only.

luan-cestari:
@bserdar OK, I will rename for the suggestion (I thought the same thing when I was adding the new field but I thought to minimize the impact of the change to not rename, but in the other hand the names there would be confusing)

bserdar:
pls also add comments to DocCtx and the getters/setters explaining that
DocCtx has three copies of the doc, and what those three copies are.

luan-cestari:
Alright

luan-cestari:
@bserdar fixed

PR #394, opened 2015-04-30
seems we would want the exception logged in this case

dcrissman:


bserdar:
"...hook or type..." : /or/of/

dcrissman:
@bserdar good catch. change made.

bserdar:
In general, pls do not reformat code unless the original is really ugly.
Generates unnecessary diffs, and I don't necessarily agree the new format
is better.

PR #400, opened 2015-05-05
Fix #214 : undeclared OR  unused Maven dependencies 

### LINKED ISSUE #214: Review the dependencies on the projects
We could use 'mvn dependency:analyze' on the projects to find not used dependencies (or with other tool). We can also checkout some dependencies with wrong scope.

luan-cestari:
Related to the issue https://github.com/lightblue-platform/lightblue-core/issues/214


bserdar:
Do not remove slf4j-simple dependencies. Without that, unit tests won't write logs.

luan-cestari:
Fixed the slf4j-simple and other problems that I found

PR #405, opened 2015-05-18
return ObjectNode instead of JsonNode, as that is what is actually being returned anyway

dcrissman:
This can save casting later on.


bserdar:
Covariant return won't save casting, as that method is used through the super class anyway. The change is harmless though.

PR #409, opened 2015-05-22
Mediator client

dcrissman:
I am finding more and more places where it would be useful to have a standard way for plugins and extensions to the lightblue framework to communicate with the Mediator directly. This PR will provide that in a new encapsulation called MediatorClient. This client can easily be injected using the MediatorClientAware interface.


dcrissman:
Dropping to in-progress as @bserdar raised some concerns about whether or not the mediator portions of this belong at the framework level because "[this] is more or less use case specific".

I might split the factory aware parts into a separate PR. I will keep this PR open for further contemplation.

dcrissman:
Will recreate when a better home can be found. Closing for now.

Issue #428, opened 2015-07-17
Clean up depreciated method in AbstractCRUDTestController


dcrissman:
This is clean up to #427.

This may have cascading effects to tests in many downstream modules.

dcrissman:
PRs are submitted. They start with https://github.com/lightblue-platform/lightblue-core/pull/429 and cascade from there.

PR #429, opened 2015-07-24
fixes: #428 - remove deprecated method

dcrissman:
This cleans up and actually simplifies the AbstractCRUDTestController. This change absolutely has cascading effects that will break builds. PRs in other repos will be linked.


dcrissman:
Other changes are actually linked to lightblue-mongo, which needs to be merged next and all else are dependent on.

PR #430, opened 2015-07-30
Synchronization extensions

bserdar:


extensions/src/main/java/com/redhat/lightblue/extensions/ExtensionSupport.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public interface ExtensionSupport {
+    /**
+     * Returns an instance of the given extension
+     *
+     * @param extensionClass The class for the requested extension
+     *
+     * @return An instance of the extension, or null if the controller doesn't support this extension
+     */
+    Extension  getExtensionInstance(Class<? extends Extension> extensionClass);
```
dcrissman: Should the returned Extension be of the same type as the extensionClass? Why not create a generic type?

bserdar: It should, but couldn't figure out the syntax to get it done using
generics. Do you have a suggestion?

dcrissman: Should be:
```java
E <E extends Extension> getExtensionInstance(Class<E> extensionClass);
```

dcrissman: Oops... other way around, sorry.

```java
<E extends Extension> E getExtensionInstance(Class<E> extensionClass);
```

bserdar: Then, when I override:

    public <E extends Extension> E getExtensionInstance(Class<E>
extensionClass) {
        if(extensionClass.equals(LockingSupport.class))
            return (Extension)new MongoLockingSupport(this);
        else
            return null;
    }

I get:

 error: incompatible types

MongoLockingSupport implements LockingSupport, which extends Extension

dcrissman: Strange, this looks good to my eyes. I pulled down your branches and maven compiled and ran the tests just fine.

bserdar: Jdk 7 or 8? 8 isn't compiling it. I think this is valid too, might be a
compiler bug.

dcrissman: Jdk 7

config/src/main/java/com/redhat/lightblue/config/LightblueFactory.java
```diff
@@ -257,6 +265,31 @@ public Object fromJson(JsonNode node) {
...
+        if(lockingMap==null) {
+            LOGGER.debug("Initializing locking");
+            Map<String,LockingSupport> map=new HashMap<>();
+            CRUDController[] controllers=getFactory().getCRUDControllers();
+            LOGGER.debug("Got {} controllers",controllers.length);
+            for(CRUDController controller:controllers) {
+                LOGGER.debug("Inspecting {}",controller.getClass());
+                if(controller instanceof ExtensionSupport) {
+                    LOGGER.debug("{} supports extensions",controller.getClass());
+                    LockingSupport lockingSupport=(LockingSupport)((ExtensionSupport)controller).getExtensionInstance(LockingSupport.class);
```
dcrissman: The LockingSupport cast should be unnecessary now.

dcrissman: I won't hold you up because of that

PR #431, opened 2015-08-04
ValueGenerators parsing and json schema changes

paterczm:
Metadata changes for https://github.com/lightblue-platform/lightblue/pull/22. Without sequences implementation in lightblue-mongo, this does nothing. It should be merged together with lightblue-mongo changes.

### DIFF
--- a/metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
+++ b/metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
@@ -1,0 +1,42 @@
+package com.redhat.lightblue.metadata;
+
+import java.io.Serializable;
+import java.util.Properties;
+
+/**
+ * Represents value generator definition located in entityInfo.
+ *
+ * @author mpatercz
+ *
+ */
+public class ValueGenerator implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    public enum ValueGeneratorType {
+        IntSequance;
+    }
+
+    private ValueGeneratorType type;
+    private String name;
+    private Properties properties = new Properties();
+
+    public ValueGenerator(ValueGeneratorType type, String name) {
+        super();
+        this.type = type;
+        this.name = name;
+    }
+
+    public ValueGeneratorType getType() {
+        return type;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public Properties getProperties() {
+        return properties;
+    }
+
+}
--- a/metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
+++ b/metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
@@ -14,21 +14,22 @@ public class ValueGenerator implements Serializable {
     private static final long serialVersionUID = 1L;
 
     public enum ValueGeneratorType {
-        IntSequance;
+        IntSequence;
     }
 
-    private ValueGeneratorType type;
-    private String name;
-    private Properties properties = new Properties();
+    // defines how values are generated
+    private final ValueGeneratorType valueGeneratorType;
+    private final String name;
+    private final Properties properties = new Properties();
 
-    public ValueGenerator(ValueGeneratorType type, String name) {
+    public ValueGenerator(ValueGeneratorType valueGeneratorType, String name) {
         super();
-        this.type = type;
+        this.valueGeneratorType = valueGeneratorType;
         this.name = name;
     }
 
-    public ValueGeneratorType getType() {
-        return type;
+    public ValueGeneratorType getValueGeneratorType() {
+        return valueGeneratorType;
     }
 
     public String getName() {
--- a/crud/pom.xml
+++ b/crud/pom.xml
@@ -49,6 +49,11 @@
           <artifactId>lightblue-core-util</artifactId>
           <version>${project.version}</version>
        </dependency>
+       <dependency>
+          <groupId>com.redhat.lightblue</groupId>
+          <artifactId>lightblue-core-extensions</artifactId>
+          <version>${project.version}</version>
+       </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
--- a/extensions/pom.xml
+++ b/extensions/pom.xml
@@ -29,6 +29,11 @@
     <packaging>jar</packaging>
     <name>lightblue-core: ${project.groupId}|${project.artifactId}</name>
     <dependencies>
+        <dependency>
+            <groupId>com.redhat.lightblue</groupId>
+            <artifactId>lightblue-core-metadata</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
     <build>
     </build>
--- a/extensions/src/main/java/com/redhat/lightblue/extensions/sequence/Sequence.java
+++ b/extensions/src/main/java/com/redhat/lightblue/extensions/sequence/Sequence.java
@@ -1,0 +1,22 @@
+package com.redhat.lightblue.extensions.sequence;
+
+import com.redhat.lightblue.metadata.SimpleField;
+
+/**
+ * Sequence generation interface, to be implemented by the backend. Can return any value, but likely
+ * an integer is going to be used.
+ *
+ * @author mpatercz
+ *
+ */
+public interface Sequence<T> {
+
+    /**
+     * Generate next value in sequence. This should be handled automatically by the backend on insert.
+     *
+     * @param field
+     * @return
+     */
+    public T nextInSequence(SimpleField field);
+
+}
--- a/extensions/src/main/java/com/redhat/lightblue/extensions/sequence/SequenceSupport.java
+++ b/extensions/src/main/java/com/redhat/lightblue/extensions/sequence/SequenceSupport.java
@@ -1,0 +1,20 @@
+package com.redhat.lightblue.extensions.sequence;
+
+import com.redhat.lightblue.extensions.Extension;
+
+/**
+ * Sequence extension provides backend specific sequence generation capabilities.
+ *
+ * @author mpatercz
+ *
+ */
+public interface SequenceSupport<T> extends Extension {
+
+    /**
+     *
+     * @param backend e.g. mongo
+     * @return
+     */
+    Sequence<T> getSequenceInstance(String backend);
+
+}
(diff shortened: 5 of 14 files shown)


paterczm:
Uid fields are generated by UIDInterceptor. I think we can use interceptor for generating other values too, relaying on a db specific extension (similar to how LockingSupport extension).

bserdar:
So: a constraint that associates a sequence name with a field, and an
interceptor that generates values on insert/update, right?

paterczm:
> So: a constraint that associates a sequence name with a field, and an interceptor that generates values on insert/update, right?

Yes, except it's not a constraint, it's a generator (new keyword).

bserdar:
Ok, saw that.

In that case, you have to change SimpleField as well, add generators to it.

paterczm:
> In that case, you have to change SimpleField as well, add generators to it.

Why SimpleField and not Field? Field has constraints, so I would think it should have value generators as well.

bserdar:
If you add to Field, then arrays and objects will have generators as well,
but not array elements.

paterczm:
Since we decided to use backend functionality to handle the sequence, I don't think the interceptor is needed. Value will be generated by the database and sequence defined by lightblue-mongo (or other backend with sequence support). For the same reason did not make any changes in LightblueFactory.

Had to add some dependencies between lightblue-core modules, pls verify.

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.redhat.lightblue.metadata.SimpleField;
+
+/**
+ * TODO: Does SimpleField have all the info required by the backend to manage sequences?
+ *
+ * @author mpatercz
+ *
+ */
+public interface Sequance {
```
bserdar: Sequence, not Sequance

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public void createSequance(SimpleField field);
+
+    public void removeSequance(SimpleField field);
+
```
bserdar: What will create and remove do?

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/SequanceSupport.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Sequence extension provides backend specific sequence number generation capabilities.
+ *
+ * @author mpatercz
+ *
+ */
+public interface SequanceSupport extends Extension {
+
+    Sequance getSequanceInstance();
+
```
bserdar: Should this be getSequenceInstance(String name) where name is datasource, or backend?

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * Generate next value in sequance. This should be handled automatically by the backend on insert.
+     *
+     * @param field
+     * @return
+     */
+    public Long nextInSequance(SimpleField field);
```
bserdar: BigInteger instead of Long?

metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ */
+public class ValueGenerator implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    public enum ValueGeneratorType {
+        IntSequance;
+    }
+
```
bserdar: Instead of this, using com.redhat.lightblue.metadata.types.Type might be better.

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
paterczm: It's for creating and removing sequences using api rather than via metadata. Could be useful for plugins, etc. Or not?

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
paterczm: How would a sequence of Objects look like? I wouldn't add it just to have to cast it to BigInteger/Long.

metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
paterczm: From what I see, com.redhat.lightblue.metadata.types are field types. Generator is not a field. com.redhat.lightblue.metadata.types.Type provides methods like to/fromJson, supportsEq, supportsOrdering - how (and why) would those be used for generator types?

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
paterczm: Agreed :)

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/SequanceSupport.java
paterczm: Updated.

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
bserdar: It is easier to not rely on a createSequence call() before using the
sequence. Sequence generator can create the sequence if it is not there.

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
bserdar: You should use a Type object to describe the type of the generated
sequence. Then you can use type.toJson(Object) to add that to a json
document.

metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
bserdar: Types are value types, not field types. You use Type to convert a value
to/from Json. We'll put the generated value to a Json document, so Type is
the correct way of doing it, you simply call type.toJson(generatedValue)

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/SequanceSupport.java
```diff
@@ -0,0 +1,20 @@
...
+import com.redhat.lightblue.extensions.Extension;
+
+/**
+ * Sequence extension provides backend specific sequence number generation capabilities.
+ *
+ * @author mpatercz
+ *
+ */
+public interface SequanceSupport extends Extension {
```
bserdar: SequenceSupport, not SequanceSupport

metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
paterczm: You mean value generator definition in entityInfo is to be handled by value generator type to/from json methods?
```
           {
                "name": "testSeq",
                "type": "IntSequance",
                "configuration": [
                    {
                        "name": "min",
                        "value": "10000"
                    },
                    {
                        "name": "max",
                        "value": "99999"
                    }
                ]
            }
```

bserdar: I don't understand the question.

In entityInfo:

{
    "name":"testSeq",
    "type":"integer"
    ....
}

The configuration part will be handled by the back end anyway.

extensions/src/main/java/com/redhat/lightblue/extensions/sequance/Sequance.java
paterczm: I don't see the create/remove sequence methods as mandatory for using
sequence value generator. I thought some sequence manipulation apis can
be useful, but since it seems confusing, I will remove them. We can add
them later if we find a need.

metadata/src/main/java/com/redhat/lightblue/metadata/ValueGenerator.java
paterczm: Disregard my previous question.

I'm having a problem grasping how a new value type can be used with generators. Generators generate values of existing types. For numeric sequences it will be BigInteger, so BigInteger.toJson(generatedValue).

paterczm: The enum I created (ValueGeneratorType.IntSequence) defines not only the type of generated value (int), but also how it is generated (sequence). Are you proposing to define the type of the generated value using com.redhat.lightblue.metadata.types.Type and having a separate generator type (sequance, random, etc.) as enum?

You never said you want a new value type, it's just what I assumed thinking about generator type. Perhaps this is the source of the confusion.

bserdar: I never said "use a new type". I said "use Type". I agree that you need an
enum to specify what type of generator it is.

So we need two things: an enum to tell what type of generator it is, like
Sequence (not Sequance), and what type of value it generates (string,
biginteger, etc).

I think the type of generator determines the data type it generates: a
sequence generator generates biginteger.

This is what I suggest:

The ValueGeneratorType specifies the generator implementation. So instead
of IntSequence, use Sequence.

The Sequence implementation should have a Type getType() method, returning
BigInteger.

Does this make sense?

Also: now that we have this, we can deprecate uid datatype, and instead use
a UID generator.

paterczm: SimpleField already has a type defined. Isn't that enough? Will it help with the implementation on the lightblue-mongo side to have com.redhat.lightblue.metadata.Type set on the generator itself?

I'm not sure if using Sequence instead of IntSequence is a good idea. Generation type heavily depends on the kind of the value being generated. Value type is not just a generation parameter, e.g. a random string will have a different implementation than random decimal. ValueGeneratorTypes like RandomDecimal, RandomString make sense to me, at least you can easily see what generators are available, rather than try out different combinations of generation type and value type.

bserdar: Ok, field has type, so it can deal with it. The generator should return an
Object in that case, and we use field.type.toJson(generatedValue) to insert
the value into the doc.

Ok with IntSequence as well.

paterczm:
Squashed commits to hide all the back and forth. Defined Sequence and SequenceSupport as generics (expecting one implementation per kind of sequence).

PR #435, opened 2015-08-21
Don't mess with timezone

bserdar:


jewzaam:
How does this impact persistence of date?  We specifically didn't want to save timestamp with time zone [1].  I assume this is the one place conversion to UTC is happening.  Maybe if I knew what problem was being solved by this it would be clearer.  Is there an issue that should be linked?

[1] http://dev.docs.lightblue.io/key_decisions/README.html#how-will-we-represent-date-in-json

bserdar:
The problem is the initial decision that we return GMT. Dates are stored as
dates, not strings, so nothing changes in terms of persistence. With this
change, date values are interpreted in the local time zone, not in GMT,
that's the only change.

So, when you save a date 2015/01/01 - 04:00:00 EDT and read it back, you'll
get 2015/01/01 - 04:00:00 EDT, not 2015/01/01-08:00:00 GMT.

dcrissman:
Does this change have a sibling in the lightblue-client for ClientConstants? We also set the timezone to GMT there.

bserdar:
No, I didn't realize that. I'll remove that one as well.

PR #437, opened 2015-08-27
Bulk

### LINKED ISSUE #424: Support querying multiple entities at one time.
We've been finding that very chatty service requests mean a slow response because of the time on the wire.  The more requests that can be made at one time is better, or just getting all the data required at one time.  Going even further would be getting data for multiple entities at one time.  Today the query language supports only one since entity and version are at the top level of the lightblue request.  A requirement with this is keeping it as a backwards compatible change.

Probably would mean moving the entity + version into the query section of the request.  Make the query include an array of queries?  And if a single query object is submitted it's the current query spec?

bserdar:
Fixes #424

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/crud/AbstractBulkJsonObject.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/AbstractBulkJsonObject.java
@@ -52,7 +52,7 @@ import com.redhat.lightblue.Request;
  */
 abstract class AbstractBulkJsonObject<T extends JsonObject> extends JsonObject {
 
-    protected final List<T> entries=new ArrayList<T>();
+    private final List<T> entries=new ArrayList<T>();
 
     /**
      * Returns all entries in the bulk object
--- a/crud/src/main/java/com/redhat/lightblue/crud/AbstractBulkJsonObject.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/AbstractBulkJsonObject.java
@@ -1,0 +1,158 @@
+/*
+ Copyright 2013 Red Hat, Inc. and/or its affiliates.
+
+ This file is part of lightblue.
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.redhat.lightblue.crud;
+
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.Collections;
+import java.util.Comparator;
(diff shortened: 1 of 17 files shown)


crud/src/main/java/com/redhat/lightblue/crud/AbstractBulkJsonObject.java
```diff
@@ -0,0 +1,158 @@
...
+ *             "op": "FIND",
+ *             "<entry>": { item }
+ *         }
+ *     ]
+ *   }
+ * </pre> 
+ */
+abstract class AbstractBulkJsonObject<T extends JsonObject> extends JsonObject {
+
+    protected final List<T> entries=new ArrayList<T>();
```
dcrissman: Does this need to be protected since getEntires() also exposes it?

bserdar: No. Will change.

crud/src/main/java/com/redhat/lightblue/crud/AbstractBulkJsonObject.java
```diff
@@ -0,0 +1,158 @@
...
+     * Parses the actual entry from the given node
+     *
+     * @param entry The entry object, an element of the entries array
+     */
+    protected abstract T parseEntry(ObjectNode entry);
+    
+    /**
+     * Parses the bulk object from the given array node
+     */
+    protected void parse(ArrayNode entriesArray) {
```
dcrissman: I see only two implementations of AbstractBulkJsonObject, and only 1 uses this functionality while the other throws a UnsupportedOperationException. Would it make more sense to just move this functionality directly to BulkRequest and not even provide it in BulkResponse? Or do you plan to have more implementations?

bserdar: I left it there in case we need to implement parse for BulkResponse.
Currently, BulkRequest is only used as an input, and BulkResponse is only
used as an output. If this changes going forward, both may end up having
parse/toJson.

crud/src/main/java/com/redhat/lightblue/crud/BulkRequest.java
```diff
@@ -0,0 +1,89 @@
...
+ *         }
+ *     ]
+ *   }
+ * </pre> 
+ */
+public class BulkRequest extends AbstractBulkJsonObject<Request> {
+
+    @Override
+    public JsonNode toJson() {
+        return toJson("requests","request");
```
dcrissman: Why not iterate over entries and call their toJson methods?

bserdar: That's what the base class method does.

dcrissman: Oh, I see it. My mistake. Disregard.

crud/src/main/java/com/redhat/lightblue/crud/AbstractBulkJsonObject.java
dcrissman: I can't think of a reason why BulkResponse would be used as an input or vice-versa.

bserdar: allright, I'll push those methods to the subclasses.

bserdar: done

PR #443, opened 2015-08-31
Generators, including current time and uuid generators

bserdar:


bserdar:
Don't merge this yet.

bserdar:
This can be reviewed/merged now

crud/src/main/java/com/redhat/lightblue/crud/valuegenerators/GeneratedFields.java
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Initializes generated fields if they are not already initialized.
+ *
+ * If a generated field is required but it is not in the document, or
+ * it is in the document with a null value, it is inserted into the
+ * document and initialized.
+ *
+ * If a generated field is not required, but it exists in the document
+ * with null value, it is initialized.
```
paterczm: It is allowed to set required field to null. It's a valid value, I'm not sure if it should trigger initialization. For example, we can have a required lastUpdateDate set to null, because it has not been updated yet. It's not a very good example, since usually lastUpdateDate matches creation date before first update, but I don't have a better one.

bserdar: The line between a field not existing, and existing with a null value is
somewhat blurred. In many places, if a field doesn't exist, getting its
value will return null. That's why I considered null and non-existence as
uninitialized.

PR #453, opened 2015-09-11
Fix #452: don't initialize array length if it is not in md

### LINKED ISSUE #452: InvalidFieldReference on an empty array of reference fields that should just be ignored
When I push an empty array of reference data, I get this error on it's size field.
This is related to https://github.com/lightblue-platform/lightblue-mongo/pull/151

```json
"errorCode":"metadata:InvalidFieldReference"
```

bserdar:


dcrissman:
I suspect this change has the possibility of breaking existing unit tests that check for a size field, but do not define the field in metadata.

PR #456, opened 2015-09-13
array elem match query thinks all fields are required, causing an unc…

bserdar:
…ontrolled recursion in composite metadata construction


metadata/src/test/java/com/redhat/lightblue/metadata/CompositeMetadataTest.java
```diff
@@ -77,6 +78,21 @@ public EntityMetadata getMetadata(Path injectionField,
         }
     }
 
+    private class NonLimitingGMD extends AbstractGetMetadata {
```
paterczm: Is that class used anywhere (other than tests)?

bserdar: No,  that's just tests.

PR #463, opened 2015-09-15
fixes #459 - selectively removes hooks from entityInfo

### LINKED ISSUE #459: When unit testing using AbstractCRUDTestController, optionally disable certain hook
Hooks can be a pain when they are not needed for a unit test. There needs to be a way to disable all or some of them for a test suite.

dcrissman:
Optionally allows test cases to selectively or entirely remove hooks from in-memory JsonNodes.


paterczm:
Reviewed, ready to merge.

PR #465, opened 2015-09-15
fixes #464 - allow datasource to be changed in-memory

### LINKED ISSUE #464: When unit testing using AbstractCRUDTestController, there needs to be a non-sustitution way to change the datasource on metadata
Currently you can set the system parameter mongo.datasource to use substitution to alter metadata that has the substitution keys in it. The problem with that is for existing production metadata, there is no way to set the datasource without modifying the json. It would be nice to be able to  just drop in metadata and not need to change it.

https://github.com/lightblue-platform/lightblue-esb-hook/blob/master/publish-hook/src/test/java/com/redhat/lightblue/hook/publish/BasePublishHookTest.java#L25

dcrissman:


paterczm:
Reviewed, ready to merge.

PR #481, opened 2015-09-25
Overhaul of query field treatment in associations: instead of keeping…

bserdar:
… absolute field names, we now keep field names with their entities, and rewrite queries relative to a node. What used to work still works, what was broken is still broken, but at least now I can fix the broken stuff, namely, reference.*.field=x search

### DIFF
--- a/crud/src/test/java/com/redhat/lightblue/assoc/RelativePathTest.java
+++ b/crud/src/test/java/com/redhat/lightblue/assoc/RelativePathTest.java
@@ -1,0 +1,36 @@
+/*
+ Copyright 2013 Red Hat, Inc. and/or its affiliates.
+
+ This file is part of lightblue.
+
+ This program is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.redhat.lightblue.assoc;
+
+import org.junit.Test;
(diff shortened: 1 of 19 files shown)


crud/src/test/java/com/redhat/lightblue/assoc/RelativePathTest.java
```diff
@@ -0,0 +1,36 @@
...
+package com.redhat.lightblue.assoc;
+
+import org.junit.Test;
+import org.junit.Assert;
+
+import com.redhat.lightblue.util.Path;
+
+public class RelativePathTest {
+    @Test
+    public void testRelativePaths() throws Exception {
```
jewzaam: Would like to see path have field different from fullPath.
And add tests with cmdPrefix non-empty.

query-api/src/main/java/com/redhat/lightblue/query/MapQueryFieldsIterator.java
```diff
@@ -0,0 +1,121 @@
...
+     * Override this method to map fields.
+     */
+    protected Path map(Path p) {
+        return p;
+    }
+    
+    @Override
+    protected QueryExpression itrValueComparisonExpression(ValueComparisonExpression q, Path context) {
+        Path p=map(applyContext(context,q.getField()));
+        if(p!=null)
```
jewzaam: With the default implementation a new expression is always created.  Could it be updated to only do that when p!=q.field?

crud/src/test/java/com/redhat/lightblue/assoc/RelativePathTest.java
bserdar: Added test cases where cmdPrefix is nonempty.

Added some comments to the function itself. The path field cannot be
different from fullPath for this function, it gets the fullpath to field,
the field as it appears in the expression, and the entity prefix containing
the field, and writes an entity relative field name, so a suffix of the
field name has to match a suffix of the fullpath.

query-api/src/main/java/com/redhat/lightblue/query/MapQueryFieldsIterator.java
bserdar: Added comments. The map function should return null if the field is not to
be mapped.

PR #491, opened 2015-10-01
Query crossing entity boundaries

bserdar:


crud/src/main/java/com/redhat/lightblue/mediator/QueryPlanNodeExecutor.java
```diff
@@ -224,23 +267,23 @@ private void execute(OperationContext ctx,
...
-            if(parents!=null) {
-                for(ChildDocReference parent:parents) {
-                    resultDoc.setParentDoc(parent.getDocument().getQueryPlanNode(),parent);
-                    parent.getChildren().add(resultDoc);
-                    LOGGER.debug("Linked parent ref:{}",parent);
+
+            if(parents!=null&&!parents.isEmpty()) {
+                for(DocReference parent:parents) {
+                    if(parent instanceof ChildDocReference) {
+                        QueryPlanNode parentNode=parent.getDocument().getQueryPlanNode();
```
jewzaam: Line "QueryPlanNode parentNode=parent.getDocument().getQueryPlanNode();" can be removed.

crud/src/main/java/com/redhat/lightblue/assoc/ResultDoc.java
```diff
@@ -132,64 +139,103 @@ public void setParentDoc(QueryPlanNode parentNode,ChildDocReference ref) {
...
+     */
+    public static void associateDocs(List<ResultDoc> parentDocs,
+                                     List<ResultDoc> childDocs,
+                                     List<Conjunct> edgeClauses,
+                                     CompositeMetadata root) {
+        // We have to find a list of childDocs for every reference in a parent doc
+        LOGGER.debug("Associating {} parent docs",parentDocs.size());
+        if(!childDocs.isEmpty()&&!parentDocs.isEmpty()) {
+            QueryPlanNode childDocsNode=childDocs.get(0).node;
+            QueryPlanNode parentDocsNode=parentDocs.get(0).node;
```
jewzaam: Is it ever the case with multiple parent or child docs that the query plan node is different for different elements in those lists?  The bindingInfo is created with the plan node from the 0th element.

crud/src/main/java/com/redhat/lightblue/mediator/QueryPlanNodeExecutor.java
bserdar: Removed.

crud/src/main/java/com/redhat/lightblue/assoc/ResultDoc.java
bserdar: Those lists are documents associated with a query plan node, meaning, each
list contains documents of one entity type only.  Added some more comments.

PR #503, opened 2015-10-13
changed json parser for array projections to accept 'projection' and 'project'

snehagunta:


query-api/src/main/resources/json-schema/projection/range.json
```diff
@@ -30,7 +30,7 @@
                 "type": "integer"
             }
         },
-        "project": {
+        "projection": {
```
bserdar: Accept both project and projection. Keep "project" section, and add "projection" with same contents

query-api/src/main/resources/json-schema/projection/match.json
```diff
@@ -25,7 +25,7 @@
         "match": {
             "$ref": "/json-schema/query/choice.json#/definitions/query"
         },
-        "project": {
+        "projection": {
```
bserdar: Accept both project and projection. Keep "project" section, and add "projection" with same contents

query-api/src/test/resources/projection/schema-test-projection-single-match-many-project.json
```diff
@@ -6,7 +6,7 @@
         "op": "$eq",
         "rfield": "someuser"
     },
-    "project": [
+    "projection": [
```
bserdar: Accept both project and projection. Keep "project" section, and add "projection" with same contents

query-api/src/test/resources/projection/schema-test-projection-single-match-single-project.json
```diff
@@ -6,7 +6,7 @@
         "op": "$eq",
         "rfield": "someuser"
     },
-    "project": {
+    "projection": {
```
bserdar: Accept both project and projection. Keep "project" section, and add "projection" with same contents

jewzaam:
@bserdar @snehagunta this change will break clients using the old "project" attribute.

jewzaam:
To be clearer, making 'projection' required instead of 'project' will mean older client requests will not be schema compliant.

bserdar:
lets make it not required. Sneha, could you remove "projection" from the
required list?

Separate PR pls.

snehagunta:
@jewzaam @bserdar  Ah..I missed that detail. Will make the change.

jewzaam:
thanks @snehagunta !

PR #509, opened 2015-10-16
Fixes #474 - Hostname on response

### LINKED ISSUE #474: Include server used (hostname) in lightblue response
Would make debugging distributed lightblue easier if we knew what host had returned the response.  Specifically when testing load balancer configurations it would be great to *know* what server was used instead of looking at logs and load balancer status etc.

BVulaj:


crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -568,4 +575,14 @@ private boolean checkQueryAccess(OperationContext ctx,QueryExpression query) {
         }
         return ret;
     }
+    
+    public static String getHostname() {
```
bserdar: You can put this in a static block, and initialize only once.

jewzaam:
JSON schema needs updated
crud/src/main/resources/json-schema/response.json

jewzaam:
Note, finding there's one other place new Response is created.
https://github.com/lightblue-platform/lightblue-mongo/blob/master/metadata/src/main/java/com/redhat/lightblue/metadata/mongo/MongoMetadata.java#L605

BVulaj:
Not sure how my search didn't find that, but good thing yours did.  Will update the JSON schema.

jewzaam:
There's a repo that is just a bunch of submodules.  Makes it easier to publish sonar stats (when it was working) and to grep for things across repos.

```
git clone git@github.com:lightblue-platform/lightblue.git
cd lightblue
git submodule foreach 'git grep "new Response" | grep -v test || true'
```

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -18,6 +18,8 @@
  */
 package com.redhat.lightblue.mediator;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
```
bserdar: Remove

core-api/src/main/java/com/redhat/lightblue/Response.java
```diff
@@ -68,6 +86,10 @@ public Response(JsonNodeFactory jsonNodeFactory) {
         this.jsonNodeFactory = jsonNodeFactory;
     }
 
+    public String getHostname() {
```
bserdar: We're not doing this now, but there's nothing preventing using the lightblue core classes on the client side. If that happens, Response will contain the hostname of the client, not the server.

Insead of this: use a mutable hostname string with getter/setters. Parse/tostring it as a regular field. Initialize if to HOSTNAME in ctor.

PR #510, opened 2015-10-19
Initialize result to false before comparing

bserdar:


jewzaam:
@bserdar for this, is there ever a case where the previous state of the QueryEvaluationContext matters when calling evaluate(QueryEvaluationContext)?  I think it doesn't since a false return from this method would mean the evaulation stops, else it continues.

bserdar:
It doesn't matter. QueryEvaluationContext is unnecessary, and I plan to
remove it.

PR #512, opened 2015-10-19
resolving merge conflicts for unbounded range

snehagunta:
Changed type of 'from' and 'to' in ArrayRangeProjector and ArrayRangeProjection to Integer from int to allow for null and 0 values for 'to'. 
Made changes to element inclusion/exclusion logic 
added tests to ProjectorTest for all 4 scenarios for range

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
@@ -143,6 +143,7 @@ public class CRUDFindRequest implements Serializable {
         if (from != null && to != null) {
             ArrayNode arr = factory.arrayNode();
             arr.add(from);
+            if(to>0)
             arr.add(to);
             node.set("range", arr);
         }
@@ -175,7 +176,7 @@ public class CRUDFindRequest implements Serializable {
                 from = x.asLong();
             }
             x = node.get("to");
-            if (x != null) {
+            if (x != null && x.asLong()>0) {
                 to = x.asLong();
             }
         }
--- a/crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
@@ -140,13 +140,15 @@ public class CRUDFindRequest implements Serializable {
         if (sort != null) {
             node.set("sort", sort.toJson());
         }
-        if (from != null && to != null) {
-            ArrayNode arr = factory.arrayNode();
-            arr.add(from);
-            if(to>0)
-            arr.add(to);
-            node.set("range", arr);
-        }
+		if (from != null && to != null) {
+			ArrayNode arr = factory.arrayNode();
+			arr.add(from);
+			if (to > 0)
+				arr.add(to);
+			else
+				arr.addNull();
+			node.set("range", arr);
+		}
     }
 
     /**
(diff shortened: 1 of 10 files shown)


query-api/src/main/java/com/redhat/lightblue/query/ArrayRangeProjection.java
```diff
@@ -25,43 +25,36 @@
...
+
+	public Integer getFrom() {
+		return this.from;
+	}
+
+	public Integer getTo() {
+		if (to == null) {
+			return null;
+		} else
+			return this.to;
```
bserdar: return to; is enough.

query-api/src/main/java/com/redhat/lightblue/query/ArrayRangeProjection.java
```diff
@@ -25,43 +25,36 @@
...
+	}
+
+	@Override
+	public JsonNode toJson() {
+		ArrayNode arr = getFactory().arrayNode();
+		arr.add(getFactory().numberNode(from)).add(getFactory().numberNode(to));
+		return ((ObjectNode) super.toJson()).set("range", arr);
```
bserdar: You need null check here, and create a nullNode if from/to is null, instead of a numberNode

crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
```diff
@@ -143,6 +143,7 @@ public void toJson(JsonNodeFactory factory, ObjectNode node) {
         if (from != null && to != null) {
             ArrayNode arr = factory.arrayNode();
             arr.add(from);
+            if(to>0)
             arr.add(to);
```
bserdar: else arr.addNull()
Otherwise range array will have one element.

crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
```diff
@@ -175,7 +176,7 @@ public void fromJson(ObjectNode node) {
                 from = x.asLong();
             }
             x = node.get("to");
-            if (x != null) {
+            if (x != null && x.asLong()>0) {
```
bserdar: It can be negative, don't x.asLong()>0

crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
```diff
@@ -140,12 +140,15 @@ public void toJson(JsonNodeFactory factory, ObjectNode node) {
...
-        if (from != null && to != null) {
-            ArrayNode arr = factory.arrayNode();
-            arr.add(from);
-            arr.add(to);
-            node.set("range", arr);
-        }
+		if (from != null && to != null) {
+			ArrayNode arr = factory.arrayNode();
+			arr.add(from);
+			if (to > 0)
```
bserdar: >= instead of >

crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
```diff
@@ -168,6 +171,7 @@ public void fromJson(ObjectNode node) {
         x = node.get("range");
         if (x instanceof ArrayNode && ((ArrayNode) x).size() == 2) {
             from = ((ArrayNode) x).get(0).asLong();
+            if(!((ArrayNode) x).get(1).isNull())
             to = ((ArrayNode) x).get(1).asLong();
```
bserdar: else to=null;

crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
snehagunta: Won't adding a null node for a negative value make the range [x, null] for [x, -y] ?

PR #514, opened 2015-10-22
Optimized projection

bserdar:
Remove metadata lookups from projection.


jewzaam:
@bserdar might want to document the Measure class but won't hold up this PR for it.

PR #525, opened 2015-11-18
Fix #524 - handle nulls

### LINKED ISSUE #524: EntityVersion.version json translation is wrong
Don't translate if null. fromJson(toJson()) results in "null" not null.

bserdar:


core-api/src/main/java/com/redhat/lightblue/Request.java
```diff
@@ -93,7 +94,8 @@ public void shallowCopyFrom(Request r) {
     public JsonNode toJson() {
         ObjectNode node = getFactory().objectNode();
         node.put("entity", entityVersion.getEntity());
-        node.put("entityVersion", entityVersion.getVersion());
+        if(entityVersion.getVersion()!=null)
```
BVulaj: Are we worried about entityVersion being null or entityVersion.version?

PR #537, opened 2015-12-08
Fixed some null pointer exceptions found when doing client development.

jewzaam:


crud/src/main/java/com/redhat/lightblue/eval/SortableItem.java
```diff
@@ -38,8 +38,10 @@ public SortableItem(JsonNode node,SortFieldInfo[] sortFields) {
         this.sortFields=sortFields;
         keyValues=new Object[sortFields.length];
         for(int i=0;i<sortFields.length;i++) {
-            JsonNode valueNode=JsonDoc.get(node,sortFields[i].getName());
-            keyValues[i]=sortFields[i].getField().getType().fromJson(valueNode);
+            if (sortFields[i] != null) {
```
bserdar: Why is there an array of SortFields with null in it?

jewzaam: Don't know, I just hit this with an internal client and fixed the two
places it broke.

bserdar: That should not be. This change will probably mask another bug by working
around it.

PR #550, opened 2015-12-22
Fix #543: Propagate errors from association retrievals

### LINKED ISSUE #543: Silent failure if the user has no access to an associated entity.
In the case of an association A -> B
If a user has permission for A but not B, and projects B.
Then it would be expected that a NoAccess error is returned to the calling code, but rather no errors are returned at all and it silently fails.

The server does log the below error...
```
2015-12-10 16:48:28,455 [hystrix-FindCommand-166] ERROR [com.redhat.lightblue.util.Error] {"objectType":"error","context":"rest/FindCommand/A/find(A:0.0.1)/find","errorCode":"mongo-crud:NoAccess","msg":"find:B"}
```

bserdar:


jewzaam:
@bserdar build fails on travis and locally

bserdar:
Fix pushed.

PR #558, opened 2016-01-06
Fix #555: parallel bulk execution

### LINKED ISSUE #555: Parallelize bulk executions

bserdar:


crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -485,21 +490,67 @@ public Response find(FindRequest req) {
...
+            if(req.getOperation()==CRUDOperation.FIND) {
+                ctx.futures[i]=executor.submit(getAsyncFind((FindRequest)req));
+            } else {
+                wait(ctx);
+                switch(req.getOperation()) {
+                case INSERT: ctx.responses[i]=insert((InsertionRequest)req);break;
+                case DELETE: ctx.responses[i]=delete((DeleteRequest)req);break;
+                case UPDATE: ctx.responses[i]=update((UpdateRequest)req);break;
+                case SAVE: ctx.responses[i]=save((SaveRequest)req);break;
+                }
```
dcrissman: I see the responses are being added to the ctx now and they were not before. Was that just a miss before? Or could you explain what is going on there?

bserdar: The 'ctx' here is something used for bulk executions only, it is not the operation context. It is here to keep the association between the future values and its corresponding response. ctx wasn't there before.

PR #560, opened 2016-01-11
when configuration file cannot be found, throw exception that is more…

dcrissman:
… meaningful than NPE


dcrissman:
Associated with https://github.com/lightblue-platform/lightblue-client/issues/239

PR #561, opened 2016-01-11
Indicate the resource that is missing instead of a NPE

dcrissman:


dcrissman:
Associated with https://github.com/lightblue-platform/lightblue-client/issues/239

PR #574, opened 2016-01-20
Fix #508: also support maxResults

### LINKED ISSUE #508: Add firstIndex and maxResults to requests
Using `range : [first, last]` can be unintuitive and can be interpreted in more ways than one.  This is why many modern frameworks and paradigms use first and max, or offset and max. For an example, just ask what the proper usage should be in #lightblue.

I suggest deprecating `range` when we do this.

Proposed snyax:
Return the first five results.
```javascript
[
  {
    "field": "*",
    "include": true,
    // deprecate range
    // "range": [ 0, 4 ],
    "firstIndex": 0,
    "maxResults": 5
  }
]
```

Return the first five results following the fifth result.
```javascript
[
  {
    "field": "*",
    "include": true,
    // deprecate range
    // "range": [ 0, 4 ],
    "firstIndex": 4,
    "maxResults": 5
  }
]
```

Return no results:
```javascript
[
  {
    "field": "*",
    "include": true,
    // deprecate range
    // "range": [ 0, 4 ],
    "firstIndex": 5,
    "maxResults": 0
  }
]
```

Return all results (maxResults is null):
```javascript
[
  {
    "field": "*",
    "include": true,
    // deprecate range
    // "range": [ 0, 4 ],
    "firstIndex": 5,
  }
]
```

Return all results (maxResults is less than 0):
```javascript
[
  {
    "field": "*",
    "include": true,
    // deprecate range
    // "range": [ 0, 4 ],
    "firstIndex": 0,
    "maxResults": -1
  }
]
```

bserdar:


crud/src/main/resources/json-schema/findRequest.json
```diff
@@ -45,11 +45,15 @@
...
+             "description": "Alternative to 'range' allowing 'from' and 'to' to be set independently."
          },
          "to": {
              "type": "integer",
-             "description": "Alternateive to 'range' allowing 'from' and 'to' to be set independently."
+             "description": "Alternative to 'range' allowing 'from' and 'to' to be set independently."
+         },
+         "maxResults": {
+             "type": "integer",
+             "description": "Alternative to 'range' allowing 'from' and 'to' to be set independently."
```
jewzaam: Description doesn't tell me that if I set "to" and "maxResults" the value of "to" is used, not "maxResults".  Could we update it to be clearer about when maxResults can be used?

jewzaam:
For docs: https://github.com/lightblue-platform/lightblue-docs-specs/pull/32

bserdar:
Clarified

PR #582, opened 2016-02-02
Fixes #575 - Case Insensitive Index Support

### LINKED ISSUE #575: Add case insensitive indexes

BVulaj:
Please do not merge yet as this requires some script additions for mongo.


BVulaj:
See: http://dev.forum.lightblue.io/Case-insensitive-search-support-td409.html

paterczm:
Will it fail gracefully when you try to make a non string field case insensitive?

BVulaj:
@paterczm We unfortunately can't verify that on index creation.  @bserdar suggested we just handle this validation at the crud level.  Checkout [Translator#populateHiddenFields](/BVulaj/lightblue-mongo/blob/206-case-insensitive-index/mongo/src/main/java/com/redhat/lightblue/mongo/crud/Translator.java#L442)

metadata/src/main/java/com/redhat/lightblue/metadata/Index.java
```diff
@@ -18,7 +18,7 @@
  */
 package com.redhat.lightblue.metadata;
 
-import com.redhat.lightblue.query.SortKey;
+import com.redhat.lightblue.query.IndexSortKey;
```
bserdar: Move this class to metadata, it is not used in query.

metadata/src/main/java/com/redhat/lightblue/metadata/Indexes.java
```diff
@@ -61,6 +63,20 @@ public boolean isEmpty() {
         return indexes.isEmpty();
     }
 
+    public boolean isCaseInsensitiveKey(Path path) {
+        return getCaseInsensitiveIndexes().stream()
+                .anyMatch(i -> i.getField().equals(path));
+
+    }
+
+    public List<IndexSortKey> getCaseInsensitiveIndexes() {
```
bserdar: What is the use of this? It returns sort keys from possibly many indexes.

metadata/src/main/java/com/redhat/lightblue/metadata/Indexes.java
```diff
@@ -61,6 +63,20 @@ public boolean isEmpty() {
         return indexes.isEmpty();
     }
 
+    public boolean isCaseInsensitiveKey(Path path) {
```
bserdar: This checks if the field is case insensitive in any of the indexes, right? The same field can appear as both case sensitive and case insensitive in multiple indexes, so what is the practical use for this?

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
```diff
@@ -306,15 +307,17 @@ public Index parseIndex(T object) {
                 }
 
                 if (null != fields && !fields.isEmpty()) {
-                    List<SortKey> f = new ArrayList<>();
+                    List<IndexSortKey> f = new ArrayList<>();
 
                     for (T s : fields) {
                         String fld = getRequiredStringProperty(s, "field");
                         String dir = getStringProperty(s, "dir");
+                        String ci = getStringProperty(s, "caseInsensitive");
```
bserdar: use getValueProperty, typecast it to Boolean.

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
```diff
@@ -306,15 +307,17 @@ public Index parseIndex(T object) {
...
+
                         if (dir == null) {
                             dir = "$asc";
                         }
-                        SortKey sort = new SortKey(new Path(fld), "$desc".equals(dir));
+                        IndexSortKey sort = new IndexSortKey(new Path(fld), "$desc".equals(dir), Boolean.valueOf(ci));
```
bserdar: ci can ne null, NPE here

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
```diff
@@ -306,15 +307,17 @@ public Index parseIndex(T object) {
...
                         f.add(sort);
```
bserdar: You need to write the caseinsensitive flag in convertIndexes

metadata/src/main/java/com/redhat/lightblue/metadata/Indexes.java
BVulaj: This is only used in one place right now, in Translator.  It's used to populate any hidden fields that have a case insensitive index.  We could just move this there if you don't see any need to have this here.

metadata/src/main/java/com/redhat/lightblue/metadata/Indexes.java
BVulaj: This is another one only used in Translator. It's just used as a convenience in case insensitive regex queries to see if we have a case insensitive field available to query.  Since it's only used there at this point, we could move it.

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
BVulaj: If ci is null here, valueOf will return 'false'.

metadata/src/main/java/com/redhat/lightblue/metadata/Indexes.java
bserdar: That is the problem with the approach that only adds methods when needed. Mongo translator might need it that way, but another translator might need it in some other way. Instead, add methods that can be used in different situations. Index.isCaseInsensitiveKey() is more meaningful than Indexes.isCaseInsensitiveKey(). You can check if the field has any case insensitive indexes over all indexes in mongo translator.

Same goes for getCaseInsensitiveIndexes() method. It is only useful from mongo translator. Move the function there, and use Index.isCaseInsensitiveKey()

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
BVulaj: Doing it that way throws a typecase exception. FYI.

bserdar: Are you using "true"/"false" in your tests? Change that to true/false, and it won't throw typecast exception.

BVulaj: Nope, the typecast exception is thrown right at this line.

Changing that to the following is enough: String cannot be case to boolean, blah, blah.

```
boolean ci = (boolean) getValueProperty(s, "caseInsensitive");
```

bserdar: It is optional though, so cast it to Boolean and check null.

PR #583, opened 2016-02-03
Jackson

dcrissman:
Upgrade jackson version and a couple other minor dependency adjustments.

### DIFF
--- a/query-api/src/test/java/com/redhat/lightblue/query/ValueTest.java
+++ b/query-api/src/test/java/com/redhat/lightblue/query/ValueTest.java
@@ -42,7 +42,6 @@ public class ValueTest {
      */
     @Test
     public void testToJson() throws IOException {
-        // if in int range returns IntNode
         Value instance = new Value(10L);
         JsonNode expResult = new LongNode(10);
         JsonNode result = instance.toJson();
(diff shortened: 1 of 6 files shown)


dcrissman:
This might have issue with our existing usage of Java 7 Dates. Going to postpone this until it can be further looked into.

query-api/src/test/java/com/redhat/lightblue/query/ValueTest.java
```diff
@@ -44,7 +44,7 @@
     public void testToJson() throws IOException {
         // if in int range returns IntNode
```
bserdar: This comment is wrong, it always returns LongNode

dcrissman:
Problem has nothing to do with Java 7 dates.

The date issue is caused from a change in the jackson library. Before if using the JsonFormat annotation on a field of a pojo, a Shape type did not need to be specified. With the upgraded version, now a Shape of type STRING needs to be provided. This is a fix that needs to be made downstream libraries.

That said, a default date format can and probably should be set. To this end, I made a change where if you create your ObjectMapper using JsonUtils the correct DateFormat will be used.

Example:
```java
@JsonFormat(pattern = DateType.DATE_FORMAT_STR, shape = Shape.STRING)
private Date creationDate;
```

PR #587, opened 2016-02-11
Honor LightblueFactoryAware hook config parsers

alechenninger:
The motivation for this is because hooks are created by parsers, and it is more natural to inject what the hook needs in its constructor (constructor injection), which simplifies testing. This requries the parser to have hook dependencies, which is why it'd be nice if parsers could get `LightblueFactory` injected.

### DIFF
--- a/config/src/main/java/com/redhat/lightblue/config/SimpleHookResolver.java
+++ b/config/src/main/java/com/redhat/lightblue/config/SimpleHookResolver.java
@@ -20,6 +20,7 @@ public class SimpleHookResolver implements HookResolver {
     public SimpleHookResolver(List<HookConfigurationParser> hookConfigurationParsers, LightblueFactory lightblueFactory) {
         if (hookConfigurationParsers != null && !hookConfigurationParsers.isEmpty()) {
             for (HookConfigurationParser parser : hookConfigurationParsers) {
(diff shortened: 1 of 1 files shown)


alechenninger:
As a side note, has anyone thought of using CDI annotations for stuff like this? You can inject a `BeanManager` from CDI and use that to create dependencies like configuration, parsers, etc, without having to worry about rolling your own dependency injection points.

alechenninger:
In other words, I could write my parser like `@Inject public MyParser(Mediator mediator)` and if you made lightblue beans available to the CDI container, that injection would "just work" if you used the `BeanManager` to create my parser class.

dcrissman:
I am under the understanding that a design decision had been made not to use a CDI. But I agree that it would be better than essentially rolling our own. Maybe now is a good time to challenge how this works.

bserdar:
Is there a way to do this using CDI without breaking everything?
Considering the amount of code needed to change this compared to the not-so-big impact, is it worth it?

alechenninger:
To be clear, this PR is not about CDI... That was just me thinking out loud
as adding manual injection points seems to be a slippery slope :). Is
everyone okay with this one line addition to inject dependencies into hook
config parsers?

PR #593, opened 2016-02-23
Method invokes inefficient Number constructor; use static valueOf instead

AymanDF:
This pull request is focused on resolving occurrences of Findbugs rule DM_NUMBER_CTOR - “ Method invokes inefficient Number constructor; use static valueOf instead ”. You can find more information about the issue here: http://findbugs.sourceforge.net/bugDescriptions.html#DM_NUMBER_CTOR
Please let me know if you have any questions.
Ayman Abdelghany.


bserdar:
Causes change in behavior

PR #596, opened 2016-02-23
Prep for mongo 216: execution options are now name-value pairs

bserdar:


jewzaam:
Schema needs updated:

crud/src/main/resources/json-schema/crudCommon.json

Probably easiest to set additionalProperties to true and be done with it since it's a map now.

jewzaam:
Working on https://github.com/lightblue-platform/lightblue-mongo/issues/219 which will benefit from this.  I'll update schema in separate PR.

jewzaam:
https://github.com/lightblue-platform/lightblue-core/pull/604

PR #600, opened 2016-03-04
Json comparator value comparison doesn't work -- could be related to …

bserdar:
…json node comparison, but fixing it anyway


util/src/main/java/com/redhat/lightblue/util/JsonCompare.java
```diff
@@ -168,4 +168,12 @@ protected JsonNode getElement(ArrayNode value,int index) {
     protected int size(ArrayNode value) {
         return value.size();
     }
+
+    public static void main(String[] args) throws Exception {
```
BVulaj: What's the purpose of having a main here?

bserdar: Testing.

PR #604, opened 2016-03-08
Updated execution section in crud json schema

jewzaam:


crud/src/main/resources/json-schema/crudCommon.json
```diff
@@ -55,17 +55,8 @@
...
-                    "description": "The upper time limit for the call in milliseconds. If the call does not complete before the limit expires, call fails with partial results written. The return status will contain what items are written."
-                },
-                "asynchronous": {
-                    "type": "integer",
-                    "description": "Given as a time limit in milliseconds. Once that limit is exceeded, the call returns, but execution continues. The return value will contain a task handle that can be used to check for execution status."
-                }
-            },
-            "additionalProperties": false
+            "description": "Execution options captured as key/value pairs.  See documentation on http://lightblue.io for what is supporeted",
+            "additionalProperties": true
```
dcrissman: Why is it preferable to leave this open ended, rather than control what properties are possible?

bserdar: It *is* open ended.

Do you suggest we define the existing properties now, and still allow
additional properties?

jewzaam: Last thing on my to-do list for adding the maxQueryTimeMS is to update user documentation.  My thinking was to just deal with this in one place, in the docs.  Especially now that execution options are passed on the operational context, there is no code (none) in core that will change if a new option is supported by a controller implementation.  It is therefore not desirable, in my opinion, to manage a concrete set of properties here if that means we do have to update the schema for those options that are controller specific.

dcrissman: Not suggesting anything, just wanted to make sure I understood. Merging.

PR #612, opened 2016-03-21
588

bserdar:


crud/src/main/java/com/redhat/lightblue/assoc/scorers/CostAndSize.java
```diff
@@ -0,0 +1,37 @@
...
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.redhat.lightblue.assoc.scorers;
+
+import java.math.BigInteger;
+
+public class CostAndSize {
+    BigInteger cost=BigInteger.ONE;
+    BigInteger size=BigInteger.ONE;
```
dcrissman: Are these suppose to be visible outside the class? If so I would suggest the final keyword with a default constructor.

bserdar: They are visible, both for read-write. It is an internal class, I can make the class package scope.

crud/src/main/java/com/redhat/lightblue/assoc/scorers/IndexedFieldScorer.java
```diff
@@ -123,53 +85,70 @@ public Comparable score(QueryPlan qp) {
...
-            
-            if(incoming==null)
-                incoming=new CostAndSize();
-            
-            // The input result set size is the join size of all incoming nodes
-            incoming.size=incoming.size.multiply( acs.size );
-            // Add up costs
-            incoming.cost=incoming.cost.add( acs.cost );
+    private BigInteger computeRetrievalCost(QueryPlanNode root) {
+        CostAndSize cs=((IndexedFieldScorerData)root.getData()).getCostAndSize();
```
dcrissman: Unchecked cast

bserdar: What's the problem with it? This class has the factory for QueryPlanData, so it expects them to be of this type.

crud/src/main/java/com/redhat/lightblue/assoc/scorers/CostAndSize.java
dcrissman: I would be fine with that

crud/src/main/java/com/redhat/lightblue/assoc/scorers/IndexedFieldScorer.java
dcrissman: But based on the method context, I don't think you can guarantee that here

bserdar: It'll throw a class cast exception if that's not the case. There is no reasonable recovery from it.

bserdar:
Ready for review. Two changes: CostAndSize is package scope, and there is type checking for node data.

PR #619, opened 2016-03-24
Add unit tests for maxResults

dcrissman:


crud/src/main/java/com/redhat/lightblue/crud/CRUDFindRequest.java
```diff
@@ -187,9 +187,12 @@ public void fromJson(ObjectNode node) {
                 x=node.get("maxResults");
                 if(x!=null) {
                     long l=x.asLong();
-                    if(l>=0) {
+                    if (l > 0) {
                         to=(from==null?0:from)+l-1;
```
dcrissman: @bserdar I am wondering if subtracting 1 is really what we want here? Setting maxResults to 0 calculates 'to' = -1. And maxResults of 1 calculates 'to' = 0.

dcrissman: @bserdar - I have done some testing and gotten some mixed results. Maybe it might be better to get some clarification.

Assuming multiple documents were already inserted. A find operation with (from:0,to:0) should return 1 jsonDoc result or 0 results? If 1 result, then what is the appropriate 'to' value to set for 0 JsonDoc results returned? -1?

bserdar: range is inclusive. [0,0] returns the 0'th document. [0,-1] returns nothing.

dcrissman:
@bserdar this is ready for review then. It just adds some unit tests.

Issue #615, opened 2016-03-23
CRUDFindRequest toJson should use prefered field names


dcrissman:
Documentation states that the range array is "Deprecated in favor of from and to / maxResults", yet CRUDFindRequest.toJson still returns the fields in this format. 

https://jewzaam.gitbooks.io/lightblue-specifications/content/language_specification/range.html

dcrissman:
Fixed in #620

PR #620, opened 2016-03-24
Insert/Update/Save now support a range

### LINKED ISSUE #614: Save/Insert/Update data operations return results, but there is no way to apply a range
My use case is when performing these operations, but I don't want the results returned back. i set the projection to exclude everything, but I still get an empty json object for each object. I still want the inserted count, but no results.

dcrissman:
Fixes #614 & #615

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/crud/withRange.java
+++ b/crud/src/main/java/com/redhat/lightblue/crud/withRange.java
@@ -1,0 +1,76 @@
+package com.redhat.lightblue.crud;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Marker interface for requests containing a range
+ */
+public interface withRange {
+
+    /**
+     * Specifies the index in the result set to start returning documents.
+     * Meaningful only if sort is given. Starts from 0.
+     */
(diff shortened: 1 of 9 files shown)


crud/src/main/java/com/redhat/lightblue/crud/withRange.java
```diff
@@ -0,0 +1,76 @@
...
+     * returned. Meaningful only if sort is given. Starts from 0.
+     */
+    Long getTo();
+
+    public static void toJson(withRange range, JsonNodeFactory factory, ObjectNode node) {
+        if (range.getFrom() != null) {
+            node.set("from", factory.numberNode(range.getFrom()));
+            if (range.getTo() != null) {
+                node.set("to", factory.numberNode(range.getTo()));
+            }
```
dcrissman: This fixes #615

crud/src/main/java/com/redhat/lightblue/crud/withRange.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Marker interface for requests containing a range
+ */
+public interface withRange {
```
bserdar: How about using capital W?

dcrissman: Fixed

PR #624, opened 2016-03-30
Metadata properties

bserdar:
Metadata parser now stores unrecognized elements as properties.

### DIFF
--- a/config/src/test/java/com/redhat/lightblue/config/TestPropertyParser.java
+++ b/config/src/test/java/com/redhat/lightblue/config/TestPropertyParser.java
@@ -22,13 +22,4 @@ import com.redhat.lightblue.metadata.parser.MetadataParser;
 import com.redhat.lightblue.metadata.parser.PropertyParser;
 
 public class TestPropertyParser extends PropertyParser {
-    @Override
-    public Object parse(String name, MetadataParser p, Object node) {
-        return null;
-    }
-
-    @Override
-    public void convert(MetadataParser p, Object emptyNode, Object object) {
-
-    }
 }
--- a/metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
+++ b/metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
@@ -67,6 +67,7 @@ import com.redhat.lightblue.metadata.Type;
 import com.redhat.lightblue.metadata.TypeResolver;
 import com.redhat.lightblue.metadata.ValueGenerator;
 import com.redhat.lightblue.metadata.Version;
+import com.redhat.lightblue.metadata.MetadataObject;
 import com.redhat.lightblue.metadata.types.ArrayType;
 import com.redhat.lightblue.metadata.types.DateType;
 import com.redhat.lightblue.metadata.types.ObjectType;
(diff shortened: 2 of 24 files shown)


config/src/test/java/com/redhat/lightblue/config/TestPropertyParser.java
```diff
@@ -22,13 +22,4 @@
...
-    @Override
-    public Object parse(String name, MetadataParser p, Object node) {
-        return null;
-    }
-
-    @Override
-    public void convert(MetadataParser p, Object emptyNode, Object object) {
-
-    }
 }
```
dcrissman: Why keep this class  around?

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
```diff
@@ -181,6 +182,16 @@ public MetadataParser(Extensions<T> ex, TypeResolver typeResolver) {
         return this.extensions;
     }
 
+    private static Set<String> asSet(String... x) {
+        HashSet<String> set=new HashSet<>();
+        for(String a:x)
+            set.add(a);
```
dcrissman: My personal preference is to use brackets even for one line loops and condition. Not a blocker to merging.

dcrissman: Why not just: new HashSet<String>(Arrays.asList(someArray));

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
```diff
@@ -1731,4 +1795,75 @@ public T getRequiredObjectProperty(T object, String name) {
      */
     public abstract void putSort(T object,String name,Sort s);
 
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+    // The above abstract methods were added based on need, and they sufffer badly from a strict top-down approach.
+    // There are multiple methods that do the same thing, and there are missing methods for crucial functionaloty.
+    // Don't use them. Use the ones below. 
+    ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
dcrissman: Should the above abstract methods be deprecated or have javadoc noting that they are not preferred? It might be useful to note what method is preferred for each.

config/src/test/java/com/redhat/lightblue/config/TestPropertyParser.java
bserdar: To test loading of a custom property parser.

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
bserdar: Done.

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
bserdar: I deprecated them. The parser still uses them,and there are other
extensions that still rely on these. Added
https://github.com/lightblue-platform/lightblue-core/issues/625

config/src/test/java/com/redhat/lightblue/config/TestPropertyParser.java
dcrissman: Why have a whole class for that when you could just as easily new PropertyParser(){}

bserdar: Extensions.getPropertyParser() returns a PropertyParser for any property
that is not recognized, and for any property for which there isn't any
registered parsers. The point of this parser is to have it in configuration
file, so we can check if the correct parser is returned for the registered
property.

Issue #655, opened 2016-05-27
Improvements to bulk processing


bserdar:
- Lets have two types of bulk requests: ordered and unordered.
  requests: [ ...requests ] : ordered request
  unorderedRequests: [ ... requests.. ]: unordered requests

unorderedRequests should be executed in parallel, based on a set parallelism count,
- For ordered requests: consecutive insert/update of the same entity should be done in a single request
- For unordered requests: all inserts/updates on the same entity should be done in a single request

paterczm:
If there are many save/insert requests, merge them into a single mongo call.

paterczm:
@bserdar, do you see a value in having both ordered and unordered arrays of requests in a bulk request? If not, maybe a simple flag - unordered: true - will suffice?

bserdar:
flag would work, should default to ordered:true to preserve existing
semantics

paterczm:
> consecutive insert/update of the same entity should be done in a single request

When you say single request, do you mean mongo bulk operation?

bserdar:
I mean a single CRUD request, which will be a single mongo bulk operation.
For instance:

bulkRequest: {
  insert entityA: { a, b, c },
  insert: entityA  { d, e, f }
}

can be
insert entityA: {a,b,c,d,e,f}

If the bulk request is unordered, all insert/save requests of the same
entity can be grouped into one. If it is ordered, only consecutive requests
can be grouped.

paterczm:
How much value is there in server side merging of insert and save crud requests? It's something the client side implementation should take care of.

bserdar:
Client side processing will not always take care of grouping requests.
There can be functions that build lightblue requests that are sometimes
called in different contexts to build a bulk request, or the request can be
submitted individually. So this optimization is useful though not essential.

PR #664, opened 2016-06-07
Unordered bulk requests

paterczm:
No request merging yet, just unordered execution in parallel.

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
+++ b/crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
@@ -580,11 +580,24 @@ public class Mediator {
         }
     }
 
-    protected Callable<Response> getAsyncFind(final FindRequest req) {
+    protected Callable<Response> getFutureRequest(final Request req) {
         return new Callable<Response>() {
             @Override
             public Response call() {
-                return find((FindRequest) req);
+                switch (req.getOperation()) {
+                    case FIND:
+                        return find((FindRequest) req);
+                    case INSERT:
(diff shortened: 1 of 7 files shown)


crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -594,33 +607,40 @@ public BulkResponse bulkRequest(BulkRequest requests) {
...
+                            case UPDATE:
+                                ctx.responses[i] = update((UpdateRequest) req);
+                                break;
+                            case SAVE:
+                                ctx.responses[i] = save((SaveRequest) req);
+                                break;
+                        }
                     }
+                } else {
+                    // unordered - do them all in parallel
```
bserdar: For this case, we can merge multiple insert/save requests on the same entity into one. Once the backend returns, we have to process the results to construct the bulk response properly.

paterczm:
This is for https://github.com/lightblue-platform/lightblue-core/issues/655

bserdar:
There is an easier way.
1) Add internalSave() internalInsert() apis that take OperationContext and
Request. The existing save and insert apis would execute a few lines, and
then call internalSave and internalInsert
 2) Use these new apis in bulk operations. They give you the
OperationContext, and that contains a List of DocCtx, the documents
operated during the operation. DocCtx contains the un-projected document
(all fields), and the projected one (the one that's returned).
 3) For getting document identities from a document, use DocIdExtractor,
and DocId. You can get an instance of DocIdExtractor from entity metadata.

PR #684, opened 2016-08-26
Maintain nulls in arrays in insert, save requests

### LINKED ISSUE #592: Consistency in how lightblue treats null values in arrays
This ticket replaces https://github.com/lightblue-platform/lightblue-mongo/issues/215

An update and insert will treat null values differently at least in an array. Uncertain at this time how a save operations handles this. This story is to investigate and make changes so that the three operations treat null array values the same.

alechenninger:
Fixes #592 

There is still the potential for filtering logic to get out of sync between update and save/insert, because save/insert docs are filtered in core before passed to controller, but all update filtering would be done in controller. Maybe filtering nulls out of doc should be left up to controller, and then controller can be internally consistent.

In any case, with this update insert, update, and saves should treat nulls in arrays consistently with mongo update at least.

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/eval/FieldComparisonEvaluator.java
+++ b/crud/src/main/java/com/redhat/lightblue/eval/FieldComparisonEvaluator.java
@@ -127,14 +127,16 @@ public class FieldComparisonEvaluator extends QueryEvaluator {
                             }
                         }
                         LOGGER.debug(" lvalue={} rvalue={}", lvalueNode, rvalueNode);
-                        if (ldocValue != null && rdocValue != null) {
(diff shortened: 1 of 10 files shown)


bserdar:
filterNulls is called on the documents coming from the client, and this should take care of preserving nulls in arrays for insert/save.

Projector.project() does another level of null filtering. Is there a test out there that validates the behavior we want? That is, do the nulls in an array preserved during projection? Is there is, then we can merge this PR, if not, pls add one.

alechenninger:
I did not see any null filtering in projector, but I did add tests, and found a bug with field comparisons where the field's value is null.

PR #710, opened 2016-10-11
709

bserdar:


crud/src/main/java/com/redhat/lightblue/assoc/Conjunct.java
```diff
@@ -54,6 +65,202 @@
...
+            }
+            throw new RuntimeException("Query processing error: cannot find clause in query field info:"+clause.toString());
+        }
+
+        private QueryFieldInfo[] findTwo(QueryExpression clause) {
+            QueryFieldInfo[] ret=new QueryFieldInfo[2];
+            int i=0;
+            for (QueryFieldInfo fi : fieldInfo) {
+                if (fi.getClause() == clause) {
+                    ret[i++]=fi;
```
BVulaj: Could this return more than two?

crud/src/main/java/com/redhat/lightblue/assoc/Conjunct.java
```diff
@@ -46,6 +48,15 @@
     private static final long serialVersionUID = 1l;
 
     /**
+     * Type of the conjunct. Value means the conjunct compares a field
+     * to a value. relation means conjunct compares fields of an
+     * entity to the fields of another entity. Comples means either
```
dcrissman: Comples - typo

bserdar:
fixed.

bserdar:
Not in the way it is used. It is only called for leaf expressions, and
those can have either one or two entities in them.

crud/src/main/java/com/redhat/lightblue/assoc/Conjunct.java
```diff
@@ -54,6 +65,202 @@
...
+                    first=false;
+                } else if(ret.type!=ConjunctType.complex) {                    
+                    ItrState qstate=iterate(subq,context);
+                    // We compare if this expression is compatible with existing assessment
+                    switch(qstate.type) {
+                    case value:
+                        if(ret.type==ConjunctType.value) {
+                            // Comparing value expression to another value expression
+                            // At most one entity can be present
+                            if(ret.numEntities==0||ret.e1==qstate.e1) {
```
dcrissman: If nothing happens, why not just negate the condition and move the else block up?

crud/src/main/java/com/redhat/lightblue/assoc/Conjunct.java
```diff
@@ -54,6 +65,202 @@
...
+                                ; // No changes, compatible expressions
```
dcrissman: Semi-colon by itself if unnecessary.

crud/src/main/java/com/redhat/lightblue/assoc/Conjunct.java
```diff
@@ -54,6 +65,202 @@
...
+            boolean first=true;
+            for(QueryExpression subq:q.getQueries()) {
+                if(first) {
+                    ret=iterate(subq,context);
+                    first=false;
+                } else if(ret.type!=ConjunctType.complex) {                    
+                    ItrState qstate=iterate(subq,context);
+                    // We compare if this expression is compatible with existing assessment
+                    switch(qstate.type) {
+                    case value:
```
dcrissman: There is a lot of logic in these case statements. Perhaps private methods might help the readability?

crud/src/main/java/com/redhat/lightblue/assoc/Conjunct.java
```diff
@@ -88,6 +295,39 @@ public QueryExpression getClause() {
...
+               clause instanceof NaryFieldRelationalExpression) {
+                // These are simple 2-term expressions
+                return ConjunctType.relation;
+            } else {
+                // At this point, clause is one of:
+                //   - ArrayMatchExpression
+                //   - NaryLogicalExpression
+                //   - UnaryLogicalExpression
+                ClauseIterator itr=new ClauseIterator();
+                return itr.iterate(clause).type;
```
dcrissman: What about the case where an unexpected Class is passed in. Should an unsupported exception be thrown?

bserdar:
done

bserdar:
Done

bserdar:
There can't be an unexpected class here. Any addition to query processing
logic affects the base class for query iterator.

BVulaj:
@dcrissman @bserdar nothing outstanding.  Merging!

Issue #334, opened 2015-03-02
Create session based id for coorelating errors


dcrissman:
When a request is made to lightblue, a uuid should be created that then all logs and responses includes. This will provide a mechanism to correlate message across errors and with clients.

Issue #324, opened 2015-02-18
Sonar - Update to the latest version


luan-cestari:
Our sonar  (sonar.lightblue.io) is on 4.3.2 and there is already an new version (5) available (http://nemo.sonarqube.org/) (http://www.sonarqube.org/downloads/). The reason for the update is to get the bug fixes and improvements (https://jira.codehaus.org/secure/ReleaseNote.jspa?projectId=11694&version=20209). This issue started from a discussion on https://github.com/lightblue-platform/lightblue-core/issues/169 .

PR #718, opened 2016-10-19
Fix #717: If there is a final filtering step, apply limits after that…

### LINKED ISSUE #717: Associations with a filtering step calculates result set size wrong
If an association execution plan has a filtering step, limits are applied before the filter, so we don't count all possible results.

bserdar:
…, not before

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/assoc/ep/ExecutionPlan.java
+++ b/crud/src/main/java/com/redhat/lightblue/assoc/ep/ExecutionPlan.java
@@ -225,11 +225,13 @@ public class ExecutionPlan {
                         if (requestSort != null) {
                             last = new Source<>(new SortResults(block, last, requestSort));
                         }
-                        if (from != null) {
-                            last = new Source<>(new Skip(block, from.intValue(), last));
-                        }
-                        if (to != null) {
-                            last = new Source<>(new Limit(block, to.intValue() - from.intValue() + 1, last));
+                        if(!needsFinalFiltering) {
+                        	if (from != null) {
+                        		last = new Source<>(new Skip(block, from.intValue(), last));
+                        	}
+                        	if (to != null) {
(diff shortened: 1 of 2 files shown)


crud/src/main/java/com/redhat/lightblue/assoc/ep/ExecutionPlan.java
```diff
@@ -225,11 +225,13 @@ public ExecutionPlan(QueryExpression requestQuery,
                         if (requestSort != null) {
                             last = new Source<>(new SortResults(block, last, requestSort));
                         }
-                        if (from != null) {
-                            last = new Source<>(new Skip(block, from.intValue(), last));
-                        }
-                        if (to != null) {
-                            last = new Source<>(new Limit(block, to.intValue() - from.intValue() + 1, last));
+                        if(!needsFinalFiltering) {
```
dcrissman: I see how this would not limit the results if additional filtering is required, but how are limits applied after filtering?

bserdar: Done in lines #324-#334, when final filter step is added

dcrissman: Is there a reason to not just apply the limit after the filter step to eliminate duplication of logic?

bserdar: Filtering will not always be there. Associations are evaluated in either one or two phases, an optional search phase, and the retrieval phase. If it is  a trivial association, there is no search phase, and everything is done in retrieval phase. But if it is not a trivial association, we first search, find the documents that match the query, and then retrieve those. Depending on the queries, it is possible that not all queries are included in the search phase, meaning it finds more docs than necessary. We add additional filtering only if all association queries cannot be evaluated at the search phase. So, duplication is necessary here. If we can limit during search, we do that. If we can't limit during search, we add a filter, and limit there.

Issue #695, opened 2016-09-20
Add support for index options


bserdar:
....to support things like partial indexes in mongo: https://github.com/lightblue-platform/lightblue-mongo/issues/311

paterczm:
Duplicate to https://github.com/lightblue-platform/lightblue-mongo/issues/311.

PR #719, opened 2016-10-19
Arbitrary index options

paterczm:
https://github.com/lightblue-platform/lightblue-core/issues/695

### DIFF
--- a/metadata/src/main/java/com/redhat/lightblue/metadata/Index.java
+++ b/metadata/src/main/java/com/redhat/lightblue/metadata/Index.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Properties;
 import java.util.Set;
 
 import com.redhat.lightblue.util.Path;
@@ -34,6 +35,7 @@ public class Index extends MetadataObject {
     private String name;
     private boolean unique = false;
     private final ArrayList<IndexSortKey> fields = new ArrayList<>();
+    private final Properties options = new Properties();
 
     /**
      * Default ctor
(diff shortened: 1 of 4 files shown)


metadata/src/main/java/com/redhat/lightblue/metadata/Index.java
```diff
@@ -135,4 +137,13 @@ public boolean isCaseInsensitiveKey(Path path) {
                 .filter(IndexSortKey::isCaseInsensitive)
                 .anyMatch(i -> i.getField().equals(path));
     }
+
+    /**
+     *
+     * @return arbitrary index options, can be backend specific
+     */
+    public Properties getOptions() {
```
bserdar: You don't really need this. Index extends MetadataObject, and it contains a Properties property. Any unrecognized property will be placed into that properties object as an object map.

metadata/src/main/java/com/redhat/lightblue/metadata/parser/MetadataParser.java
```diff
@@ -328,6 +328,16 @@ public Index parseIndex(T object) {
                 }
                 parseProperties(index, object, INDEX_ELEMENTS);
 
+                // parse arbitrary index options, can be backend specific
+                T options = getMapProperty(object, "options");
```
bserdar: Similarly, you shouldn't need this., Any unrecognized property will be in 'properties'

paterczm:
What is the difference between https://github.com/lightblue-platform/lightblue-mongo/issues/311 and https://github.com/lightblue-platform/lightblue-core/issues/695, if any? I guess I can just close the issue in core.

bserdar:
I believe they're the same

PR #722, opened 2016-11-01
Fixes #715 association cycles

### LINKED ISSUE #715: Association cycles using default versions is broken if entity occurs in the cycle with a given version
A -> A, association uses default version, but A is requested using a given version. Response is crud:MetadataAppearsTwice.

There is a question here: If A->A association uses the default version, and if A is loaded using a given version that is not the same as the default version, does the relationship use the given version, or the default version?

If the answer is "default version", then A->A relationship is only valid when we load the default version of A.

If the answer is "given version", then A->A defines a separate relation for each version of A.

I think associations using default versions should be evaluated using the given version of an entity. That is, if A -> A is done using default version D, and if A is read using a version V!=D, then the association uses version V, not D.

BVulaj:


crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -67,16 +68,24 @@
...
-            return DefaultMetadataResolver.this.getMetadata(entityName, entityVersion);
+        protected EntityMetadata retrieveMetadata(Path injectionPath, String entityName, String entityVersion) {
+            EntityMetadata emd = metadataMap.get(entityName);
+            if (emd != null) {
+                String defaultVersion = emd.getEntityInfo().getDefaultVersion();
+                if (!isValidReference(entityVersion, requestVersion, defaultVersion)) {
+                    throw Error.get(CrudConstants.ERR_METADATA_APPEARS_TWICE, entityName + ":" + entityVersion + " and " + requestVersion);
+                }
+            }
+            return metadataMap.get(entityName);
```
bserdar: return emd;

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +191,81 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
+        Set<String> deferred = new HashSet<>();
+        FieldCursor cursor = emd.getFieldCursor();
+        while (cursor.next()) {
+            FieldTreeNode node = cursor.getCurrentNode();
+            if (node instanceof ReferenceField) {
+                ReferenceField ref = (ReferenceField) node;
+                String name = ref.getEntityName();
+                String ver = ref.getVersionValue();
+                EntityMetadata refMd;
+                if (ver == emd.getEntityInfo().getDefaultVersion()) {
```
bserdar: if association is using default version, ver=null. So, this comparison is wrong. if(ver==null), then add this to deferred loads, so you can load it later.

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +191,81 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
+                    refMd = metadataMap.get(entityName);
+                } else {
+                    refMd = md.getEntityMetadata(name, ver);
```
bserdar: If ver!=null, then association requires a definite version. If a different version of the entity is already used, give error. If default version of the entity is referenced before, then that default version should be this version.

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +191,81 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
+     *
+     * @param version
+     *            The version of a referenced entity
+     * @param requestVersion
+     *            The version of the root entity that has been requested
+     * @param defaultVersion
+     *            The default version of the root entity
+     * @return
+     */
+    private boolean isValidReference(String version, String requestVersion, String defaultVersion) {
```
bserdar: This isn't used anymore

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +191,81 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
-        EntityMetadata emd = metadataMap.get(entityName);
-        if (emd != null) {
-            if (!emd.getVersion().getValue().equals(version)) {
-                throw Error.get(CrudConstants.ERR_METADATA_APPEARS_TWICE, entityName + ":" + version
-                        + " and " + emd.getVersion().getValue());
+    private void initMetadataMap(EntityMetadata emd, String entityName, String rootRequestVersion) {
+        // store root entity metadata
+        metadataMap.put(entityName, emd);
+        Set<String> deferred = new HashSet<>();
+        FieldCursor cursor = emd.getFieldCursor();
```
bserdar: You need to separate this into a different recursive function, and descend down the entity tree. This only processes the root level entity.

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +185,54 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
-            if (emd == null || emd.getEntitySchema() == null) {
-                throw Error.get(CrudConstants.ERR_UNKNOWN_ENTITY, entityName + ":" + version);
+    private void initMetadataMap(EntityMetadata emd, String entityName, String rootRequestVersion) {
+        // store root entity metadata
+        metadataMap.put(entityName, emd);
+        Set<String> deferred = new HashSet<>();
+        FieldCursor cursor = emd.getFieldCursor();
+        while (cursor.next()) {
+            FieldTreeNode node = cursor.getCurrentNode();
+            validateNode(emd, entityName, rootRequestVersion, deferred, node);
```
bserdar: Cursor traverses field tree in depth first manner. call validateNode only if field is instanceof ReferenceField

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +185,54 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
+        }
+
+        // we didn't find specific versions for these entities, so use default
+        deferred.forEach(n -> {
+            String def = md.getEntityInfo(n).getDefaultVersion();
+            metadataMap.put(n, md.getEntityMetadata(n, def));
+        });
+    }
+
+    private void validateNode(EntityMetadata emd, String entityName, String rootRequestVersion, Set<String> deferred, FieldTreeNode node) {
```
bserdar: Rename this. We are not validating nodes.

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +185,54 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
+        if (node instanceof ObjectField) {
```
bserdar: Don't do this, cursor traverses all fields of the entity.

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
bserdar: Instead of FieldTreeNode, use   ReferenceField arg

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +185,54 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
+                deferred.add(name);
+            } else { // else, explicit version
+                refMd = metadataMap.get(name);
+                if (!refMd.getVersion().getValue().equals(ver)) { // if we already have a different version loaded
+                    throw Error.get(CrudConstants.ERR_METADATA_APPEARS_TWICE, name + ":" + ver + " and " + refMd.getVersion().getValue());
+                } else if (deferred.contains(name)) { // already used with a default version, so use the explicit version
+                    refMd = md.getEntityMetadata(name, ver);
+                    metadataMap.put(name, refMd);
+                    deferred.remove(name);
+                }
```
bserdar: else // if it is not deferred
 { metadataMap.put(name,refMd) }

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -180,25 +185,54 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
+        FieldCursor cursor = emd.getFieldCursor();
+        while (cursor.next()) {
+            FieldTreeNode node = cursor.getCurrentNode();
+            validateNode(emd, entityName, rootRequestVersion, deferred, node);
+        }
+
+        // we didn't find specific versions for these entities, so use default
+        deferred.forEach(n -> {
+            String def = md.getEntityInfo(n).getDefaultVersion();
+            metadataMap.put(n, md.getEntityMetadata(n, def));
```
bserdar: Check if default version exists for this entity, throw exception if not

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
BVulaj: I thought the cursor wouldn't traverse down into other objects?  Or did I misunderstand that?  Will they traverse everything but referenced objects?

bserdar:
The cursor will traverse all fields, depth first.

I think the source of your confusion is how reference fields are handled.
When you load metadata from the db, it has ReferenceField objects for
entity references. These are leaf objects, they only give the name/version
of the entity, and the query. The cursor we have in this part of the code
is iterating over the fields of such metadata objects. So, if entity E is
defined as:

E {
    field1: {string},
    field2: { object,
           fields: {
                x:{string},
                ref:{reference to entity F}
           }
   }

The cursor will visit field1, field2, x, ref. 'ref' is a ReferenceField.

The composite metadata for E will have 'ref' replaces with a
ResolvedReferenceField. A ResolvedReferenceField extends an
ObjectArrayField (so 'ref' is converted to an array), and the object
definition of each element is built from the fields of entity F. This way,
composite metadata will have fields from all references entities as well.

So:

A field cursor you get from composite metadata will visit all fields of all
entities in that composite metadata.

A field cursor you get from metadata will visit all fields of that entity.

PR #734, opened 2016-11-18
fix #732: rewrite metadata loading, remove old query analyzer

### LINKED ISSUE #732: 3-level projection problem
A->B->C, project C, and nothing comes.

bserdar:
Fixes #732, rewrites recursive metadata loading algorithm, removed old query analyzer


crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -183,43 +185,60 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
...
     
-    private void validateRefField(EntityMetadata emd, String entityName, String rootRequestVersion, Set<String> deferred, ReferenceField ref) {
+    private void validateRefField(EntityMetadata emd, Deque<EntityMetadata> toProcess, Set<String> deferred, ReferenceField ref) {
         String name = ref.getEntityName();
         String ver = ref.getVersionValue();
-        EntityMetadata refMd = null;
+        EntityMetadata refMd = metadataMap.get(name);
         if (ver == null) { // if null, default version
-            deferred.add(name);
+            if(refMd==null)
```
dcrissman: Could/Should this consolidated into a single condition? if((ver == null) && (refMd == null))

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
```diff
@@ -183,43 +185,60 @@ private void addFieldRoles(Set<String> roles, FieldTreeNode node) {
         }
     }
 
-    private void initMetadataMap(EntityMetadata emd, String entityName, String rootRequestVersion) {
-        // store root entity metadata
-        metadataMap.put(entityName, emd);
+    private void initMetadataMap(EntityMetadata emd, String rootEntityName, String rootRequestVersion) {
```
bserdar: emd is non-null at this point, a null emd will fail with null ptr exceptions later.

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
bserdar: We defer loading if ver==null and it is not already loaded. There is an else to (if ver==null) part where the actual loading happens, that's why this can't be consolidated into a single condition.

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
dcrissman: Thanks for the clarification

crud/src/main/java/com/redhat/lightblue/mediator/DefaultMetadataResolver.java
dcrissman: Just double checking. Thanks for the clarification.

PR #744, opened 2016-12-05
Add tools directory, md2swagger

bserdar:


tools/md2swagger.py
```diff
@@ -0,0 +1,103 @@
...
+        properties['format']='date-time'
+    elif type=='double':
+        properties['type']='number'
+        properties['format']='double'
+    elif type=='integer':
+        properties['type']='integer'
+        properties['format']='int64'
+    elif type=='string':
+        properties['type']='string'
+    return properties;
```
dcrissman: how about a default case if unexpected type is passed in

tools/md2swagger.py
```diff
@@ -0,0 +1,103 @@
...
+
+def singular(s):
+    if s[-3:]=='ies':
+        return s[0:-3]+'y'
+    elif s[-4:]=='sses':
+        return s[0:-4]+'ss'
+    elif s[-1]=='s':
+        return s[0:-1]
+    else:
+        return s;
```
dcrissman: Seems frail

bserdar:
I'll add the default case for unknown type (string).

Do you have a better idea to convert a plural to a singular?

tools/md2swagger.py
```diff
@@ -0,0 +1,103 @@
...
+                fieldBlock['items']={'$ref':'#definitions/'+newEntityName}
+                toplevel[newEntityName]=processFieldTree(fieldContents['fields'],toplevel,nameStack)
+                nameStack.pop()
+            elif fieldContents['type']=='array':
+                fieldBlock['type']='array'
+                arrayType=fieldContents['items']['type']
+                if arrayType=='object':
+                    nameStack.append(singular(capital1(field)))
+                    newEntityName=makeEntityName(nameStack)
+                    fieldBlock['items']={'$ref':'#definitions/'+newEntityName}
```
alechenninger: I made some fixes in the internal repo for this you may want to grab
- need #/definitions (missing /)
- The new top level sub entities need type: object and properties: fields

tools/md2swagger.py
dcrissman: > Do you have a better idea to convert a plural to a singular?

In fairness, no. My thought was that a word that correctly ends with 's' would have problems, but the more I think about it, that may be a fringe case anyway.

Could you explain the need for this conversion?

bserdar:
Without this, we had UserLegalEntities as a classname.

tools/md2swagger.py
dcrissman: > Without this, we had UserLegalEntities as a classname.

I won't hold this up further on this point. This is certainly good enough and we can adjust if/when we find a use case. Thanks.

bserdar:
Added fixes from the internal repo, and added a catch-all for type
properties.

PR #756, opened 2017-01-19
Saved search

bserdar:


crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -91,8 +91,8 @@
 
     private static final Path OBJECT_TYPE_PATH = new Path("objectType");
 
-    private final Metadata metadata;
-    private final Factory factory;
+    public final Metadata metadata;
+    public final Factory factory;
```
bvulaj: Necessary?

misc/src/main/java/com/redhat/lightblue/savedsearch/FindRequestBuilder.java
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * Builds a find request from the saved search by rewriting the
+     * search bound parameters using parameter values. The search
+     * query components can be either strings, or JSON objects. If
+     * they are strings, the parameters are replaced first, then the
+     * resulting json string is converted to a query/projection. If
+     * the components are JSON objects, then the values of those json
+     * objects are replaced with parameter values.
+     */
+    public static FindRequest buildRequest(JsonNode savedSearch,
```
bvulaj: What's the workflow for the user here?  Get the saved search from SavedSearchCache and then pass it in here with their params?

misc/src/main/java/com/redhat/lightblue/savedsearch/SavedSearchCache.java
```diff
@@ -0,0 +1,388 @@
...
+     *
+     * @param m Mediator instance
+     * @param clid The client id
+     * @param searchName name of the saved search
+     * @param entit Name of the entity
+     * @param version Entity version the search should run on
+     *
+     * @return The saved search document, or null if not found
+     */
+    public JsonNode getSavedSearch(Mediator m,
```
bvulaj: Can we name these a little more intuitively?  `getSavedSearch` and `getSavedSearchFromDb` or something like that.

misc/src/main/metadata/lb-saved-search.json
```diff
@@ -0,0 +1,169 @@
...
+                        }
+                    }
+                }
+            },
+            "query":{
+                "type":"any",
+                "constraints":{
+                    "required":true
+                },
+                "description":"The query. Bound parameters are given as ${parameter}. This could be a strig, or json "
```
bvulaj: typo

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
bserdar: Used in getSavedSearch. No need to hide these as they're final, either add getters, or make public. This seems more straight forward. Without exposing these, I have to go into configuration to get an instance of Metadata, and that's introducing cyclic dependency,

misc/src/main/java/com/redhat/lightblue/savedsearch/FindRequestBuilder.java
bserdar: The implementation is in rest, I'll push the PR once this is in. The flow is: parse URL, get saved search from the cache, build request using url params, and run the search.

misc/src/main/metadata/lb-saved-search.json
bserdar: fixed

misc/src/main/java/com/redhat/lightblue/savedsearch/FindRequestBuilder.java
bserdar: Added more documentation describing the workflow.

misc/src/main/metadata/lb-saved-search.json
```diff
@@ -0,0 +1,169 @@
...
+            "entity": {
+                "type":"string",
+                "constraints":{
+                    "required":true
+                },
+                "description":"The name of the entity search works on"
+            },
+            "versions": {
+                "type":"array",
+                "description": "Entity versions this search can be applied to. It is a partial version match. A 1.0 value in this array means the search will be used for all 1.0.x versions as well as 1.0 version. A null value means the search will be used for all entity versions that are not explicitly selected by other elements in the array.",
```
paterczm: When will the default entity version be used?

bserdar:
In the most intuitive way, and after several iterations, I decided that it
is this:

If you call:
     /search/getById/user?id=1
i.e. no version: This uses the default version of user. Whatever that
default is, if there is a saved search for that specific version, that will
be used. Otherwise, saved search containing null version will be used.

If you call:
   /search/getById/user/1.0.0?id=1
This will use the user:1.0.0 version. If there is a search for 1.0.0, 1.0,
or 1, that will be used. Otherwise, search containing null version will be
used.

PR #757, opened 2017-01-23
Doc metadata

bserdar:
Expose docversion in response. Optionally inject docversion in the doc (that part is done in mongo).


crud/src/test/java/com/redhat/lightblue/mediator/BulkTest.java
```diff
@@ -221,26 +221,26 @@ public void parallelOrderedBulkTest() throws Exception {
             @Override
             public void run() {
                 try {
-                    // Check if all 3 finds are waiting
+                    System.out.println("Check if all 3 finds are waiting");
```
bvulaj: I would use log statements instead if possible.

bserdar: Fixed. Also fixed the test so it won't hang every other run.

PR #761, opened 2017-01-26
Update if same

bserdar:


crud/src/main/java/com/redhat/lightblue/crud/CRUDOperationContext.java
```diff
@@ -53,6 +53,8 @@
     private final CRUDOperation CRUDOperation;
     private final HookManager hookManager;
     private final ExecutionOptions executionOptions;
+    private final Set<String> documentVersions=new HashSet<>();
+    private boolean updateIfSame;
```
paterczm: I would call it updateIfCurrent or updateIfNotStale. Same is not as clear to me.

crud/src/main/resources/json-schema/response.json
```diff
@@ -69,7 +91,7 @@
                 }
             }
         },
-        "otherErrors": {
+        "errors": {
```
paterczm: What is this change about?

crud/src/main/java/com/redhat/lightblue/crud/CRUDOperationContext.java
bserdar: Done.

crud/src/main/resources/json-schema/response.json
bserdar: This was wrong already. It is "errors", not "otherErrors"

PR #764, opened 2017-02-21
Streams

bserdar:
Stream result sets instead of passing them around in lists. This is the first step for dealing with large results without running out of memory.


crud/src/main/java/com/redhat/lightblue/assoc/CompositeFindImpl.java
```diff
@@ -140,7 +144,9 @@ public void explain(OperationContext ctx,
         initialize(ctx,req);
         ExecutionContext executionContext = new ExecutionContext(ctx,null);
         JsonDoc doc=new JsonDoc(executionPlan.explain(executionContext));
-        ctx.addDocument(doc);        
+        ArrayList<DocCtx> l=new ArrayList<>(1);
+        l.add(new DocCtx(doc));
+        ctx.setDocumentStream(new ListDocumentStream<DocCtx>(l));
```
bserdar: done

crud/src/main/java/com/redhat/lightblue/assoc/ep/StepResultDocumentStream.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Iterator;
+import java.util.ArrayList;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+import com.redhat.lightblue.crud.DocCtx;
+import com.redhat.lightblue.crud.DocumentStream;
+
+/**
+ * Interface between DocumentStream used in CRUDOperationContext, and StepResult
```
bserdar: Not java interface, but an interface between two different streams

crud/src/main/java/com/redhat/lightblue/crud/DocumentStream.java
```diff
@@ -0,0 +1,92 @@
...
+public interface DocumentStream<T> extends Iterator<T>{
+    /**
+     * Close the document stream
+     */
+    void close();
+
+    /**
+     * Adds a listener that will be called for each document of the stream when next() is called
+     */
+    void tee(Consumer<T> dest);
```
bserdar: tee, as in pipes, and unix tee

crud/src/main/java/com/redhat/lightblue/crud/DocumentStream.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * returns true if rewind works
+     */
+    default boolean canRewind() { return false; }
+
+    /**
+     * returns a new stream that starts the same resultset from the beginning. Only works if canRewind() is true
+     */
+    default DocumentStream<T> rewind() { throw new UnsupportedOperationException(); }
```
bserdar: Added RewindableDocumentStream

crud/src/main/java/com/redhat/lightblue/crud/DocumentStream.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public static <S,D> DocumentStream<D> map(final DocumentStream<S> source,final Function<S,D> map) {
```
bserdar: done

crud/src/main/java/com/redhat/lightblue/crud/DocumentStream.java
```diff
@@ -0,0 +1,92 @@
...
+        return new DocumentStreamMapper<S,D>(source,map);
+    }
+
+    static final class DocumentStreamMapper<S,D> implements DocumentStream<D> {
```
bserdar: not final any more. Why separate file?

crud/src/main/java/com/redhat/lightblue/crud/DocumentStream.java
dcrissman: I still don't think 'tee' is a very good method name.

bserdar:
Fine, renamed. But that's exactly what it does:

https://en.wikipedia.org/wiki/Tee_(command)

crud/src/main/java/com/redhat/lightblue/crud/DocumentStream.java
```diff
@@ -0,0 +1,77 @@
...
+public interface DocumentStream<T> extends Iterator<T>{
+    /**
+     * Close the document stream
+     */
+    void close();
+
+    /**
+     * Adds a listener that will be called for each document of the stream when next() is called
+     */
+    void forEach(Consumer<T> dest);
```
dcrissman: What about add() or append()?

bserdar:
Like addListener()?

But this isn't exactly a listener. It is a function that will be called for
each doc, as they are pulled from the stream, so forEach tells what it does

bserdar:
Maybe I can change the comment ?

crud/src/main/java/com/redhat/lightblue/crud/DocumentStream.java
dcrissman: > But this isn't exactly a listener. It is a function that will be called for each doc, as they are pulled from the stream, so forEach tells what it does

Maybe I need to see an example of the method being used? I don't see any calls in this PR, and the only implementations I see just add the Consumer to an array called listeners. I don't even think I see a place that any action is being performed on those arrays.

bserdar:
https://github.com/lightblue-platform/lightblue-core/pull/764/files#diff-1a122660679e1163deac1643961a046fR282

crud/src/main/java/com/redhat/lightblue/crud/DocumentStream.java
dcrissman: > I don't even think I see a place that any action is being performed on those arrays.

I take that back, I see it now.

crud/src/main/java/com/redhat/lightblue/hooks/HookManager.java
```diff
@@ -90,25 +92,71 @@ public DocHooks(DocCtx doc, Map<Hook, CRUDHook> hooks) {
...
-    private static final class HookDocs {
-        private final Hook hook;
-        private final CRUDHook crudHook;
-        private final EntityMetadata md;
-        private final List<HookDoc> docs = new ArrayList<>();
+    private class HookAndDocs {
+        final Hook hook;
+        final EntityMetadata md;
+        final CRUDHook resolvedHook;
+        final List<HookDocInfo> docList=new ArrayList<>();;
```
bserdar: fixed

bserdar: Renamed forEach to addListener

PR #776, opened 2017-04-11
move up from ldap controller

dcrissman:
I have long intended to move these classes from the ldap controller to core. The intention of these classes are to make translating from native objects to json easier and provide a standard mechanism for handling lightblue fields that may not be able to be stored in the underlying data store and need to be generated (eg. objectType and array# fields).

The is an initial attempt to move the files over. I am open to suggestions for improvements and namespacing.


dcrissman:
If/when merged, I will remove from the ldap controller.

crud/src/main/java/com/redhat/lightblue/crud/LightblueUtil.java
```diff
@@ -0,0 +1,113 @@
...
+import com.redhat.lightblue.metadata.ArrayField;
+import com.redhat.lightblue.metadata.Field;
+import com.redhat.lightblue.metadata.Fields;
+
+/**
+ * Utility methods to assist with interacting with the Lightblue framework.
+ *
+ * @author dcrissman
+ */
+public final class LightblueUtil {
```
bserdar: Contents of this class can be moved to metadata/.../PredefinedFields.java

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
```diff
@@ -0,0 +1,185 @@
...
+        fieldCursor.parent();
+        return valueNode;
+    }
+
+    protected abstract JsonNode translate(ReferenceField field, Object o);
+    protected abstract JsonNode translate(SimpleField field, Object o);
+
+    protected abstract Object getValueFor(S source, Path path);
+    protected abstract List<? extends Object> getSimpleArrayValues(Object o);
+    protected abstract List<? extends Object> getObjectArrayValues(Object o);
```
bserdar: This is probably not sufficient, the implementation will need the field cursor.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * Defines a class that take a response from a datasource and translates it
+ * into something Lightblue can use.
+ *
+ * @author dcrissman
+ *
+ * @param <S> - The source type that this {@link TranslatorToJson} converts
+ * to json.
+ */
+public abstract class TranslatorToJson<S> {
```
bserdar: There are two functions in util/.../JsonUtil, that translate a json doc to a list/map tree and back. Take a look at that. It doesn't use metadata, so this is an extension of that.

dcrissman:
Pushing up what I have, but more work to be done.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
dcrissman: I assume you are referring to valueToJson and valueFromJson. These do look similar to the Translators. The main difference that I can see is that these methods are converting a single value, whereas the Translators are converting a specific class back and forth from json. I use it in LDAP to convert json to/from native library classes.

Might the JsonUtils methods be a simple sub-type of the translator hierarchy?

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
dcrissman: Are you just highlighting getObjectArrayValues? Or also getSimpleArrayValues and getValueFor?
For just 'ToJson'? or also 'FromJson'?

I assume your concern is just for ObjectArrayValues, for which I do not have any concrete implementations that don't just throw an unsupported. So I expect the cursor is needed for the object fields? Is that what you are thinking?

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
bserdar: I am talking about JsonUtils.fromJson and JsonUtils.toJson. They translate a list/map object tree to/from json.

These classes translate json using metadata so those two may not be compatible, so disregard what I said, There is not much to share between the two.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
bserdar: getObjectArrayValues will need the cursor to get the metadata fields, and parse the object entries. Without that, there is no way it can know what fields it needs to parse.

getSimpleArrayValues needs at least the SimpleArrayElement object, so it can know the type of the value it is processing.

In general, you need to psss the cursor down to these methods, so the implementation can traverse the metadata tree. However, there should be a strict convention on how the cursor should be used, so it might make sense to provide an overridable body for these methods that iterate through the cursor.

dcrissman: > getObjectArrayValues will need the cursor to get the metadata fields, and parse the object entries. Without that, there is no way it can know what fields it needs to parse.

I don't actually have a use case for this scenario, but I think I can take a stab at it.

> getSimpleArrayValues needs at least the SimpleArrayElement object, so it can know the type of the value it is processing.

What about just passing in the Type itself?

bserdar: Passing in the SimpleArrayElement would be better.

dcrissman:
@bserdar - Changes made. Any other suggestions?

I still need to write some unit tests for this. Especially for the new logic. Right now the only tests for it live in the LDAP controller for it's specific use cases.

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
```diff
@@ -0,0 +1,70 @@
...
+    public NonPersistedPredefinedFieldTranslatorFromJson(EntityMetadata entityMetadata) {
+        super(entityMetadata);
+    }
+
+    @Override
+    protected void translate(JsonNodeCursor cursor, T target){
+        Path path = cursor.getCurrentPath();
+        FieldTreeNode fieldNode = entityMetadata.resolve(path);
+        if(fieldNode instanceof Field){
+            String datasourceFieldName = ((Field) fieldNode).getName();
```
bserdar: datasourceFieldName is misleading, just fieldName?

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
```diff
@@ -0,0 +1,70 @@
...
+
+            Fields fields = entityMetadata.getFields();
+            if(fieldNode instanceof ObjectField){
+                fields = ((ObjectField) fieldNode).getFields();
+            }
+
+            if(PredefinedFields.isFieldObjectType(datasourceFieldName)
```
bserdar: How about adding a new function PredefinedFields.isPredefinedField(fieldName,md) ?

metadata/src/main/java/com/redhat/lightblue/metadata/PredefinedFields.java
```diff
@@ -217,6 +213,82 @@ private static ParentNewChild ensureArraySize(EntityMetadata md, ArrayField arr)
         return ret;
     }
 
+    /**
+     * Returns <code>true</code> if the passed in field name is the object type, otherwise <code>false</code>.
+     * @param fieldName - field name to test.
+     * @return <code>true</code> if the passed in field name is the object type, otherwise <code>false</code>.
+     */
+    public static boolean isFieldObjectType(String fieldName){
+        if(fieldName == null){
```
bserdar: No need for null check

metadata/src/main/java/com/redhat/lightblue/metadata/PredefinedFields.java
```diff
@@ -217,6 +213,82 @@ private static ParentNewChild ensureArraySize(EntityMetadata md, ArrayField arr)
...
+        return false;
+    }
+
+    /**
+     * Creates and returns the array field name for the passed in array count field name.
+     * @param countField - array count field.
+     * @return the array field name for the passed in array count field name.
+     */
+    public static String createArrayFieldNameFromCountField(String countField){
+        if(countField == null){
```
bserdar: Just a single line is enough:

if(doesFieldName...) then return countField.substring(...) else countField

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    void appendToJsonNode(S source, ObjectNode targetNode, FieldCursor fieldCursor){
+        targetNode.set(fieldCursor.getCurrentNode().getName(), translate(source, fieldCursor));
+    }
+
+    private JsonNode translate(S source, FieldCursor fieldCursor) {
+        FieldTreeNode field = fieldCursor.getCurrentNode();
+
+        Error.push(field.getFullPath().getLast());
```
bserdar: This is expensive. field.getName() would do.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
```diff
@@ -0,0 +1,203 @@
...
+
+        try{
+            Object value = getValueFor(source, fieldCursor.getCurrentPath());
```
bserdar: How does this work under an array? If cursor is iterating under, say, arr.*.x, then there are many 'x' fields under the array arr. 

It looks like it would work, but I can't help but thinking that this is more complicated than it needs to be,

dcrissman: It would require the implementing class to know what value to return from the source based on the path. I think a unit test for this would come in handy.

metadata/src/main/java/com/redhat/lightblue/metadata/PredefinedFields.java
dcrissman: Not sure what you mean by only needing a single line, but the null check was redundant with doesFieldNameMatchArrayCountPattern.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
dcrissman: Thinking more about this, getValueFor would return the entire array/collection, not just a single element in it. It would then be the the translate(ArrayField) methods job to know how to extract each child item.

dcrissman: I think that I am beginning to see your problem. I am certainly open to suggestions.

The issue that I was attempting to solve with getValueFor is that only the concrete implementation will know the logic to extract a value from a source object that isn't a pojo. This seems to work great on flat objects with simple fields (which is more of less all that LDAP needs), but not as well for an embedded object scenario.

bserdar: One option is to rewrite this so that it iterates the json doc, instead of iterating the metadata.

dcrissman: But in this case we don't have a json object, that is what we are creating. We are converting a java object to json.

Here is an example of this in action. In this example the ldap library has returned a SearchResultEntry, which then needs to be converted to json the lightblue dsl understands.
https://github.com/lightblue-platform/lightblue-ldap/blob/master/lightblue-ldap-crud/src/main/java/com/redhat/lightblue/crud/ldap/translator/ResultTranslatorToJson.java

dcrissman: The other thing is that the way the data looks in the data store may not be how it is represented in the metadata. Again, ldap for example, is basically a flat datastore, but the ldap controller allows the metadata to reorganize the fields in some meaningful way. So for that, we need to parse the metadata.

dcrissman:
Changes made to support arrays of objects in TranslatorToJson as well as unit tests. I plan yet to look at TranslatorFromJson to see if I can add the same support and unit tests.

dcrissman:
@bserdar - this is ready for another review. Now both TranslatorToJson and TranslatorFromJson support ObjectFields and ObjectArrayElements.

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
```diff
@@ -0,0 +1,69 @@
...
+public abstract class NonPersistedPredefinedFieldTranslatorFromJson<T> extends TranslatorFromJson<T>{
+
+    public NonPersistedPredefinedFieldTranslatorFromJson(EntityMetadata entityMetadata) {
+        super(entityMetadata);
+    }
+
+    @Override
+    protected void translate(JsonNodeCursor cursor, Object target) {
+        Path path = cursor.getCurrentPath();
+        FieldTreeNode fieldNode = entityMetadata.resolve(path);
```
bserdar: This throws exception if path cannot be resolved. Ignore that exception, it meansdoc has a field that's not in metadata.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
```diff
@@ -0,0 +1,212 @@
...
+                newJsonNode = translateToObjectNode(value, cursor);
+            }
+            else if((newValue = getValueFor(value, cursor.getCurrentPath())) != null){
+                if (field instanceof SimpleField) {
+                    newJsonNode = translate((SimpleField)field, newValue);
+                }
+                else if (field instanceof ArrayField){
+                    newJsonNode = translateToArrayNode((ArrayField) field, newValue, cursor);
+                }
+                else if (field instanceof ReferenceField) {
```
bserdar: IIRC, ReferenceFields are converted to ResolvedReferences once they're processed. Also, a ResolvedReference extends an ArrayField, so you have to first compare if it is a resolved reference, and then if it is an arrayfield. 

Or, don't even bother. Once processed, and ReferenceField will be an ArrayField.

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
dcrissman: I assume you mean MetadataConstants.ERR_INVALID_FIELD_REFERENCE? If so, then I will log a WARN and ignore.

dcrissman: Presumably, the same will be needed here [1] as well?

[1] https://github.com/lightblue-platform/lightblue-core/pull/776/files#diff-db67fec7430644a03c1bce6a203b18e9R100

bserdar:
Don't. Just catch and ignore.

bserdar:
yes.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
dcrissman: Honestly, I wasn't sure how to handle ReferenceField(s), so for now the concrete implementations just have noop implementations.

> Or, don't even bother.  Once processed, and ReferenceField will be an ArrayField.

Are you suggesting that I just drop support for this field type? Or that I treat it like an ArrayField? ObjectArrayField maybe?

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
dcrissman: > Don't. Just catch and ignore.

Why? Wouldn't it be useful to know a transformation is incomplete?

dcrissman:
@bserdar - Currently I have these classes in the lightblue-core-crud module, but I am wondering if lightblue-core-metadata would be a more appropriate home?

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
bserdar: No, if it is an invalid field, it'll fail somewhere else. It is likely that this is a field that's not in current metadata.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
bserdar: Remove this if statement for reference fields, and ArrayField will handle it.

bserdar:
Maybe move them to metadata/.../translators ?

dcrissman:
> Maybe move them to metadata/.../translators ?

So we don't loose ongoing discussions, I will move them once everything else is approved.

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
dcrissman: If it will fail some place else, then what is the harm in leaving the code as is? This would prevent additional complex conditional logic that might not always be relevant and likely to be forgotten about.

crud/src/main/java/com/redhat/lightblue/crud/TranslatorToJson.java
dcrissman: Removed ReferenceField logic altogether as it isn't relevant without this conditional anyway.

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
bserdar: It produces unnecessary logs. Let it fail somewhere else.

dcrissman: I guess it seems odd to me to catch an unchecked exception for no reason other than to silence it. Is this maybe a compensation for an architecture issue elsewhere?

bserdar:
The decision to throw an exception from metadata.resolve() was wrong.

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
dcrissman: > The decision to throw an exception from metadata.resolve() was wrong.

I guess I don't want to hold this PR up, but would it make more sense to fix this in (it looks like) Fields.resolve(Path, int)?

bserdar:
It would make sense to fix it. We need to figure out if we have code that
relies on that exception before trying to fix it.

crud/src/main/java/com/redhat/lightblue/crud/NonPersistedPredefinedFieldTranslatorFromJson.java
dcrissman: > It would make sense to fix it. We need to figure out if we have code that relies on that exception before trying to fix it.

I created #786 to address that issue. If it is acceptable to you, I'd like to move forward with this PR as is (pending moving to metadata namespace).

dcrissman:
Ready to be reviewed and merged.

Issue #783, opened 2017-05-18
Setting referenced fields to null on update results in NPE


paterczm:
```
java.lang.NullPointerException
	at com.redhat.lightblue.eval.SetExpressionEvaluator$FieldData.access$100(SetExpressionEvaluator.java:60)
	at com.redhat.lightblue.eval.SetExpressionEvaluator.update(SetExpressionEvaluator.java:201)
	at com.redhat.lightblue.mongo.crud.IterateAndUpdate.updateDoc(IterateAndUpdate.java:287)
	at com.redhat.lightblue.mongo.crud.IterateAndUpdate.update(IterateAndUpdate.java:193)
	at com.redhat.lightblue.mongo.crud.MongoCRUDController.update(MongoCRUDController.java:350)
	at com.redhat.lightblue.mediator.Mediator.update(Mediator.java:286)
```

Update has additional concurrency control logic and can be thus preferred over save. Clients can do this: ```update.updates(Update.updatePojo(pojo));```, which can result in references being set to null.

PR #784, opened 2017-05-18
Fix for #783? Unit test needed.

paterczm:
A hint on what the problem with https://github.com/lightblue-platform/lightblue-core/issues/783 might be. Didn't test it.

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/eval/SetExpressionEvaluator.java
+++ b/crud/src/main/java/com/redhat/lightblue/eval/SetExpressionEvaluator.java
@@ -139,7 +139,9 @@ public class SetExpressionEvaluator extends Updater {
             } else if (mdNode instanceof ArrayField) {
                 data = initializeArray(rvalue, refMdNode, mdNode, field, refPath);
             }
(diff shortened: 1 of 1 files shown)


crud/src/main/java/com/redhat/lightblue/eval/SetExpressionEvaluator.java
```diff
@@ -139,7 +139,9 @@ public SetExpressionEvaluator(JsonNodeFactory factory, FieldTreeNode context, Se
             } else if (mdNode instanceof ArrayField) {
                 data = initializeArray(rvalue, refMdNode, mdNode, field, refPath);
             }
-            setValues.add(data);
+            if (data != null) {
```
bserdar: this would work, but we should also add an if statement *before* instanceof ArrayField (becayse ResolvedReferenceField extends ArrayField)

if(mdNode instanceof ResolvedReferenceField) {
    // then, if the set-to value is null, set data to null, so it won't be added to setValues. If it is non-null, throw error, because the call is setting a reference to a non-null value.
}

PR #790, opened 2017-06-07
WIP: Changes for Monitoring and Alerting (health check using dropwizard-metrics)

sameer31:


sameer31:
@alechenninger, @derek63, @svkm102, Kindly review the changes

crud/src/test/java/com/redhat/lightblue/crud/test/TestCRUDController.java
```diff
@@ -85,4 +86,10 @@ public MetadataListener getMetadataListener() {
     public void updatePredefinedFields(CRUDOperationContext ctx, JsonDoc doc) {
     }
 
+	@Override
+	public LightblueHealth checkHealth() {
```
paterczm: The checkHealth logic can be taken from https://github.com/lightblue-platform/lightblue-healthcheck.

alechenninger:
Marek, that module is from client perspective. It causes side effects. This
is a different approach that should be more scalable and provide better
insight. Haven't had a chance to review Sameers changes yet however.

crud/src/main/java/com/redhat/lightblue/crud/LightblueHealth.java
```diff
@@ -0,0 +1,32 @@
+package com.redhat.lightblue.crud;
+
+/**
+ *
+ */
+public class LightblueHealth {
```
alechenninger: I think `CRUDHealth` is probably a better name for this–it continues the pattern of the rest of the return types in CRUDController, so this is more discoverable / less surprising in my opinion.

crud/src/main/java/com/redhat/lightblue/crud/LightblueHealth.java
```diff
@@ -0,0 +1,32 @@
+package com.redhat.lightblue.crud;
+
+/**
+ *
+ */
```
alechenninger: Please add documentation or remove the javadoc comment.

crud/src/main/java/com/redhat/lightblue/crud/LightblueHealth.java
```diff
@@ -0,0 +1,32 @@
...
+public class LightblueHealth {
+	private final boolean isHealthy;
+	private final String details;
+
+	/**
+	 * @param isHealthy
+	 * @param details
+	 */
```
alechenninger: Please add documentation to the params or remove the javadoc comment

crud/src/main/java/com/redhat/lightblue/crud/LightblueHealth.java
```diff
@@ -0,0 +1,32 @@
...
+	public LightblueHealth(boolean isHealthy, String details) {
+		this.isHealthy = isHealthy;
+		this.details = details;
+	}
+
+	/**
+	 * @return if the application is healthy
+	 */
```
alechenninger: Return doc is not really helpful (no useful information), and in fact is kind of misleading–the health check isn't for an entire application. It's for a specific controller. So, either this should include useful information–like perhaps some guidelines for what "healthy" means about a controller–or we should remove the javadoc.

crud/src/test/java/com/redhat/lightblue/crud/test/TestCRUDController.java
```diff
@@ -85,4 +86,10 @@ public MetadataListener getMetadataListener() {
...
+		// TODO Auto-generated method stub
+		return null;
```
alechenninger: For this test controller, consider returning always healthy.

crud/src/test/java/com/redhat/lightblue/mediator/MockCrudController.java
```diff
@@ -107,4 +108,10 @@ public void updatePredefinedFields(CRUDOperationContext ctx, JsonDoc doc) {
         }
     }
 
+	@Override
+	public LightblueHealth checkHealth() {
+		// TODO Auto-generated method stub
+		return null;
```
alechenninger: Same, consider always healthy instead of null

crud/src/main/java/com/redhat/lightblue/crud/CRUDController.java
```diff
@@ -125,4 +120,9 @@ CRUDFindResponse find(CRUDOperationContext ctx,
      * The back end should update the predefined fields of the document
      */
     void updatePredefinedFields(CRUDOperationContext ctx, JsonDoc doc);
+    
+    /**
+     * Checks the CRUD health
```
alechenninger: Maybe make comment more meaningful or remove it. If a comment is literally the same thing as the method name it's just a waste.

crud/src/main/java/com/redhat/lightblue/crud/CRUDHealth.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public boolean isHealthy() {
+        return isHealthy;
+    }
+
+    public String details() {
+        return details;
+    }
+}
```
alechenninger: No newline at EOF–also might be a good idea to add equals/hashcode/toString.

sameer31: Accepted, would add implementations for equals/hashcode/toString

crud/src/main/java/com/redhat/lightblue/crud/CRUDController.java
sameer31: Would below comment be suitable?

"Checks the CRUD controller health. The respective implementations should provide the health of lightblue CRUD layer which is further exposed to external systems by dropwizard metrics HealthCheckServlet REST endpoint"

alechenninger:
Looks good! Let's wait to merge until mongo and ldap are ready so we don't break the build for too long.

PR #792, opened 2017-06-08
Poor man's profiling

paterczm:
A very simple and unintrusive logging of crud operations which take more than 5 seconds, in lieu of proper profiling instrumentation. Main purpose is to be able to see inside bulk operations.

Here is how a log line looks:

```
[main] WARN stopwatch.com.redhat.lightblue.mediator.Mediator - #find({"entity":"A","entityVersion":"1.0.0","op":"FIND","query":{"field":"obj1.c.*.field1","op":"$eq","rvalue"
:"ABFPwrjyx-o5DQWWZmSEfKf3W1z"},"projection":[{"field":"*","include":true,"recursive":true} in 3.04ms
```

There is no way to configure this without recompilation (except configuring the logger to stop logging this).

### DIFF
--- a/crud/pom.xml
+++ b/crud/pom.xml
@@ -65,6 +65,11 @@
             <artifactId>json-schema-core</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.jcabi</groupId>
+            <artifactId>jcabi-aspects</artifactId>
+            <version>0.22.6</version>
+        </dependency>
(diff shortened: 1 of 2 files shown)


paterczm:
This solution has "smart time" feature (https://github.com/jcabi/jcabi-aspects/issues/222), which changes time unit depending on the duration. Not parsing friendly.

PR #794, opened 2017-06-09
Memindexconfig

paterczm:
Make memoryIndexThreshold configurable.
* < 0 means never use in memory index
* = 0 means always use in memory index
* \> 0 means use in memory index when number of results to associate is > memoryIndexThreshold

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/assoc/ep/Assemble.java
+++ b/crud/src/main/java/com/redhat/lightblue/assoc/ep/Assemble.java
@@ -218,7 +218,7 @@ public class Assemble extends Step<ResultDocument> {
                 }
                 // Try to build an index from results
                 MemDocIndex docIndex=null;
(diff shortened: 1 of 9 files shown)


crud/src/main/java/com/redhat/lightblue/assoc/ep/Assemble.java
```diff
@@ -216,7 +218,7 @@ public void commit() {
                 }
                 // Try to build an index from results
                 MemDocIndex docIndex=null;
-                if(aq.getQuery()!=null&&numSlots>MEM_INDEX_THRESHOLD) {
+                if(aq.getQuery()!=null&&memoryIndexThreshold>=0&&numSlots>memoryIndexThreshold) {
```
bserdar: unnecessary >0 check

paterczm: I remembered why this check is useful - you can disable this feature with memoryIndexThreshold=-1. Without it you'd need to set memoryIndexThreshold to some very large number.

PR #795, opened 2017-06-09
Poor man's profiling - custom solution

paterczm:
A very simple and unintrusive logging of crud operations, in lieu of proper profiling instrumentation. Main purpose is to be able to see inside bulk operations, but can be easily added to other lightblue modules and methods.

Warning: StopWatches on mediator operations will log entire requests, including secrets. For that reason this feature is disabled by default. It can be controlled using system properties (-Dstopwatch.enabled=true to start logging, -Dstopwatch.globalWarnThresholdMS=10000 to change global threshold).

### DIFF
--- a/util/src/main/java/com/redhat/lightblue/util/JsonUtils.java
+++ b/util/src/main/java/com/redhat/lightblue/util/JsonUtils.java
@@ -443,6 +443,33 @@ public final class JsonUtils {
         }
     }
 
+    /**
+     * An approximation of how much space will a JsonNode take in memory. Does not take jvm optimizations
+     * or architecture into account. Useful only to compare JsonNodes with each other.
+     *
+     * @param node
+     * @return
+     */
+    public static int size(JsonNode node) {
+        int size = 0;
+        if (node instanceof ArrayNode) {
+            for (Iterator<JsonNode> elements = ((ArrayNode) node).elements(); elements.hasNext();) {
+                size += size(elements.next());
+            }
+        } else if (node instanceof ObjectNode) {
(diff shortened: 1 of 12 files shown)


paterczm:
This is similar to https://github.com/lightblue-platform/lightblue-core/pull/792, except it offers some key improvements:
* can be disabled
* logging threshold can be changed without recompiling code
* will log time always in milliseconds, will not change units

paterczm:
Added logic to check result size and log it if it exceeds threshold.

util/src/main/java/com/redhat/lightblue/util/JsonUtils.java
```diff
@@ -443,6 +443,33 @@ private static void indent(StringBuilder bld, int depth) {
...
+        } else if (node instanceof ObjectNode) {
+            for (Iterator<Map.Entry<String, JsonNode>> fields = ((ObjectNode) node).fields(); fields.hasNext();) {
+                Map.Entry<String, JsonNode> field = fields.next();
+                size += field.getKey().length();
+                size += size(field.getValue());
+            }
+        } else if (node instanceof NumericNode) {
+            size += 4;
+        } else {
+            size += node.asText().length();
```
bserdar: Need nullcheck here

paterczm: Can a node ever be null (except when null is passed to JsonUtils.size)?

bserdar: No, but null can be passed to JsonUtils.size() in this context, and this is the only place it'll throw a NPE. Null was not a concern where this was copied from.

bserdar:
+ return JsonUtils.size(response.getEntityData());

response.getEntityData() may be null here.

PR #798, opened 2017-06-22
Streaming mediator

bserdar:
* remove hook processing from find
 * Add new findAndSteam API to stream results

### DIFF
--- a/core-api/src/main/java/com/redhat/lightblue/BaseResponse.java
+++ b/core-api/src/main/java/com/redhat/lightblue/BaseResponse.java
@@ -55,8 +55,8 @@ public class BaseResponse extends JsonObject {
     private String taskHandle;
     private SessionInfo session;
     private String hostname;
-    private final List<DataError> dataErrors = new ArrayList<>();
-    private final List<Error> errors = new ArrayList<>();
+    private final List<DataError> dataErrors;
+    private final List<Error> errors;
 
     final JsonNodeFactory jsonNodeFactory;
 
@@ -80,12 +80,27 @@ public class BaseResponse extends JsonObject {
      */
     @Deprecated
     public BaseResponse() {
-        jsonNodeFactory = JsonNodeFactory.withExactBigDecimals(true);
+        this(JsonNodeFactory.withExactBigDecimals(true));
     }
 
     public BaseResponse(JsonNodeFactory jsonNodeFactory) {
         this.jsonNodeFactory = jsonNodeFactory;
         this.hostname = HOSTNAME;
+        this.dataErrors=new ArrayList<>();
+        this.errors=new ArrayList<>();
+    }
+
+    public BaseResponse(BaseResponse r) {
+        jsonNodeFactory=r.jsonNodeFactory;
+        entity=r.entity;
+        status=r.status;
+        modifiedCount=r.modifiedCount;
+        matchCount=r.matchCount;
+        taskHandle=r.taskHandle;
+        session=r.session;
+        hostname=r.hostname;
+        dataErrors = r.dataErrors;
+        errors = r.errors;
     }
 
     public EntityVersion getEntity() {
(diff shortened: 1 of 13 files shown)


dcrissman:
> remove hook processing from find
What was the motivation for this change? While I agree, find probably takes the least benefit from this, it is a change to how we have done hooks.

bserdar:
With streaming, it is no longer possible to call hooks with the list of all
documents.

core-api/src/main/java/com/redhat/lightblue/BaseResponse.java
```diff
@@ -0,0 +1,236 @@
...
+
+        }
+        HOSTNAME = hostName;
+    }
+
+    /**
+     * @deprecated use Response(JsonNodeFactory)
+     */
+    @Deprecated
+    public BaseResponse() {
```
dcrissman: A new constructor that is already deprecated?

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -477,81 +477,119 @@ private QueryExpression rewriteUpdateQueryForCompositeSearch(CompositeMetadata m
...
+            factory.getInterceptors().callInterceptors(InterceptPoint.PRE_MEDIATOR_FIND, ctx);
+            Finder finder;
+            if (ctx.isSimple()) {
+                LOGGER.debug("Simple entity");
+                finder = new SimpleFindImpl(md, factory);
+            } else {
+                LOGGER.debug("Composite entity");
+                finder = new CompositeFindImpl(md);
+                // This can be read from a configuration
+                ((CompositeFindImpl) finder).setParallelism(9);
```
dcrissman: Should this be configurable?

core-api/src/main/java/com/redhat/lightblue/BaseResponse.java
bserdar: Pushed down from Response. I moved everything in Response except entityData and entityMetadata to BaseResponse, and this moved as well.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
bserdar: It should, but not a priority.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -477,81 +477,119 @@ private QueryExpression rewriteUpdateQueryForCompositeSearch(CompositeMetadata m
...
+                ctx.measure.end("find");
+                METRICS.debug("find: {}",ctx.measure);
             }
+            Error.pop();
+        }
+        
+        return response;
+    }
+    
+    @StopWatch(loggerName = "stopwatch.com.redhat.lightblue.mediator.Mediator", sizeCalculatorClass = "com.redhat.lightblue.mediator.ResponsePayloadSizeCalculator")
```
paterczm: A note on StopWatch: since findAndStream execution time will depend on the client (how fast it's consuming the results), long execution times may be fine. That's the whole point I guess, just need to keep that in mind when looking at slow queries logged.

bserdar:
findAndStream will return immediately with a cursor. The first pull from
that cursor will run the query. So, logging this particular api is probably
useless. Should we remove stopwatch?

paterczm:
> findAndStream will return immediately with a cursor. The first pull from that cursor will run the query. So, logging this particular api is probably useless. Should we remove stopwatch?

If it returns immediately, then yes, let's remove StopWatch. Better move it somewhere though. Maybe rest layer has a method which represents processing of entire stream (until it's closed)?

bserdar:
Removed. The inner streaming class in FindCommand maybe is the place to do
it, but then, its speed is dependent on the client pulling it.

PR #801, opened 2017-07-18
Result set size cap to prevent OoM errors

paterczm:
Adding logic to Mediator#find which checks the size of results list as it's being populated from a stream. If the size of that list exceeds a configured threshold, client receives crud:ResultSizeTooLarge error and no data. This is to prevent OoM errors when large documents are being returned.

It's also possible to define a warning threshold for result set size. It differs from StopWatch annotation in that it will log a warning even when Mediator#find does not return (e.g. system error, or node restarted because of OoM).

### DIFF
--- a/core-api/src/main/java/com/redhat/lightblue/Response.java
+++ b/core-api/src/main/java/com/redhat/lightblue/Response.java
@@ -49,7 +49,7 @@ public class Response extends BaseResponse  {
     private Request forRequest;
     private boolean warnThresholdBreached = false;
 
-    // TODO: Response has no access to CrudConstants
+    // Response has no access to CrudConstants
     public static final String ERR_RESULT_SIZE_TOO_LARGE = "crud:ResultSizeTooLarge";
 
 
--- a/core-api/src/main/java/com/redhat/lightblue/Response.java
+++ b/core-api/src/main/java/com/redhat/lightblue/Response.java
@@ -18,25 +18,38 @@
  */
 package com.redhat.lightblue;
 
+import java.util.Iterator;
+import java.util.List;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
+import com.redhat.lightblue.util.Error;
+import com.redhat.lightblue.util.JsonUtils;
 
 /**
  * Response information from mediator APIs
  */
 public class Response extends BaseResponse  {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(Response.class);
+
     private static final String PROPERTY_PROCESSED = "processed";
     private static final String PROPERTY_RESULT_METADATA = "resultMetadata";
 
     private JsonNode entityData;
     private List<ResultMetadata> resultMetadata;
+
+    private int responseDataSizeB = 0;
+    private int maxResultSetSizeB = -1, warnResultSetSizeB = -1;
+    private Request forRequest;
+
+    // TODO: Response has no access to CrudConstants
+    public static final String ERR_RESULT_SIZE_TOO_LARGE = "crud:ResultSizeTooLarge";
 
 
     /**
@@ -59,6 +72,24 @@ public class Response extends BaseResponse  {
 
     public Response(BaseResponse r) {
         super(r);
+    }
+
+    public void ensureResponseSizeNotTooLarge(int maxResultSetSizeB, int warnResultSetSizeB, Request forRequest) {
+        this.forRequest = forRequest;
+        this.maxResultSetSizeB = maxResultSetSizeB;
+        this.warnResultSetSizeB = warnResultSetSizeB;
+    }
+
+    public boolean isEnsureResposneSizeNotTooLarge() {
+        return maxResultSetSizeB > 0;
+    }
+
+    public boolean isWarnResponseSizeLarge() {
+        return warnResultSetSizeB > 0;
+    }
+
+    public boolean isCheckResponseSize() {
+        return isEnsureResposneSizeNotTooLarge() || isWarnResponseSizeLarge();
     }
 
     /**
@@ -87,9 +118,12 @@ public class Response extends BaseResponse  {
      */
     public void addEntityData(JsonNode doc) {
         if(doc!=null) {
+
+            enforceResponseSizeLimits(doc);
+
             if(entityData==null) {
                 entityData=JsonNodeFactory.instance.arrayNode();
-            } 
+            }
             if(doc instanceof ArrayNode) {
                 for(Iterator<JsonNode> itr=doc.elements();itr.hasNext();) {
                     ((ArrayNode)entityData).add(itr.next());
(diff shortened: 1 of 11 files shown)


alechenninger:
How hard would this be to add to update also @paterczm ?

paterczm:
> How hard would this be to add to update also @paterczm ?

Added. Also planning to add a check for bulk. At least ordered bulk should be fairly easy.

core-api/src/main/java/com/redhat/lightblue/Response.java
```diff
@@ -101,6 +135,37 @@ public void addEntityData(JsonNode doc) {
...
+     * This method is used to checks only the data part of the response. TODO: check dataErrors as well, as they include entire docs.
+     *
+     * @param entityDataJson json being added to the response
+     */
+    private void enforceResponseSizeLimits(JsonNode entityDataJson) {
+
+        if (isCheckResponseSize()) {
+            responseDataSizeB += JsonUtils.size(entityDataJson);
+
+            // TODO: could account for copies made for hooks here
```
paterczm: What do you think about multiplying the size to account for the copies made for hooks (the number of copies depends on the request kind)? Consider this a workaround until the real solution: https://github.com/lightblue-platform/lightblue-mongo/issues/378

paterczm: Decided to duplicate this logic in HookManager. It's only until https://github.com/lightblue-platform/lightblue-core/issues/802.

crud/src/test/java/com/redhat/lightblue/mediator/MediatorTest.java
```diff
@@ -401,6 +400,72 @@ public void queryTimeoutTest() throws Exception {
     }
 
     @Test
+    public void findResultSetTooLargeTest() throws Exception {
```
alechenninger: Can we add a test for update/insert/delete when there are no hooks registered?

Might even want to test even with hooks, to ensure the hookManager result size checks are setup correctly.

core-api/src/main/java/com/redhat/lightblue/Response.java
```diff
@@ -18,26 +18,40 @@
...
 
     private JsonNode entityData;
     private List<ResultMetadata> resultMetadata;
 
+    private int responseDataSizeB = 0;
+    private int maxResultSetSizeB = -1, warnResultSetSizeB = -1;
+    private Request forRequest;
+    private boolean warnThresholdBreached = false;
+
+    // Response has no access to CrudConstants
```
paterczm: I was thinking to extend core-api's Response in core-crud as CappedResponse, but decided it does not help readability.

core-api/src/main/java/com/redhat/lightblue/Response.java
```diff
@@ -61,6 +75,24 @@ public Response(BaseResponse r) {
...
+    public boolean isEnsureResposneSizeNotTooLarge() {
+        return maxResultSetSizeB > 0;
+    }
+
+    public boolean isWarnResponseSizeLarge() {
+        return warnResultSetSizeB > 0;
+    }
+
+    public boolean isCheckResponseSize() {
+        return isEnsureResposneSizeNotTooLarge() || isWarnResponseSizeLarge();
```
paterczm: Went for isErrorOnResposneSizeTooLarge and isWarnOnResponseSizeLarge.

crud/src/main/java/com/redhat/lightblue/hooks/HookManager.java
```diff
@@ -173,6 +220,10 @@ public HookManager(HookResolver r, JsonNodeFactory factory) {
      */
     public void clear() {
         queuedHooks.clear();
+
+        queuedHooksSizeB = 0;
+        forRequest = null;
+        warnThresholdBreached = false;
```
paterczm: Those values never change on runtime.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -301,6 +301,7 @@ public Response update(UpdateRequest req) {
                         updateResponse.setNumMatched(0);
                     }
                 }
+                ctx.getHookManager().ensureQueuedHooksSizeNotTooLarge(factory.getMaxResultSetSizeB(), factory.getWarnResultSetSizeB(), req);
```
paterczm: Take a look at Mediator#setResponseResults:840-843. If no hooks were queued for processing, it will use the thresholds in the Response. Mediator#setResponseResults is called for inserts, saves and deletes (all writes). Find doesn't do hooks at all, so it's always using Response implementation.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -647,22 +651,12 @@ public Response explain(FindRequest req) {
...
-        }
-    }
-
     protected void wait(BulkExecutionContext ctx) {
         for (int i = 0; i < ctx.futures.length; i++) {
             if (ctx.futures[i] != null) {
                 try {
                     LOGGER.debug("Waiting for a find request to complete");
-                    ctx.responses[i] = ctx.futures[i].get();
+                    ctx.setResponseAt(i, ctx.futures[i].get());
```
paterczm: > It would be interesting if we could allow responses to share an abstraction like a SizeCap or something of the sort. Then, we could have on SizeCap that all parallel responses shared, and then we wouldn't have to wait for them to each finish before summing their results (where it's likely some have already completed and we could already be well over the max by that time).

That's true. Unordered bulk requests are still dangerous, because all requests are executed in parallel and the limit is not effectively enforced. But nobody uses them afaik. For ordered bulk requests this should work reasonably well.

> Take this a step further ...
and use client side streaming :)

> As an aside, do we have any way to manage the size of the servlet-serving thread pool that the app server will use?

Yes we do, http executor's pool can be configured in threads subsystem of EAP. Joins are part of response, so they are included in the response size check. The problem I see is that we can introduce another bottleneck this way. At the very least this would require some tuning to get it right.

crud/src/test/java/com/redhat/lightblue/mediator/MediatorTest.java
paterczm: > Can we add a test for update/insert/delete when there are no hooks registered?
There is one already: MediatorTest#updateResultSetTooLargeTest. No save and delete, but they use the same code path for this.

> Might even want to test even with hooks, to ensure the hookManager result size checks are setup correctly.

I agree, the components/units are tested in isolation, but their integration is not tested. On the other hand the setup logic is really simple.

core-api/src/main/java/com/redhat/lightblue/Response.java
alechenninger: Can we make the error constants more visible?

If it's high effort for some reason, maybe not worth it right now.

core-api/src/main/java/com/redhat/lightblue/Response.java
```diff
@@ -61,6 +75,24 @@ public Response(BaseResponse r) {
         super(r);
     }
 
+    public void ensureResponseSizeNotTooLarge(int maxResultSetSizeB, int warnResultSetSizeB, Request forRequest) {
+        this.forRequest = forRequest;
+        this.maxResultSetSizeB = maxResultSetSizeB;
+        this.warnResultSetSizeB = warnResultSetSizeB;
+    }
```
alechenninger: The sentences that make up the method names are literally synonymous. At a glance, their arguments tell more about how they are different than their names. Using something like "set thresholds" is very clear that this merely configures the thresholds.

It's not a huge deal but it seemed like a simple readability improvement.

core-api/src/main/java/com/redhat/lightblue/Response.java
paterczm: I don't see a clean way to do this right now. CrudConstants belong to core-crud, but Response belongs to core-api. I added crud logic to Response, which goes against existing organization. The only clean way I see is creating a new CappedResponse class in core-crud (basing off of Response in core-api), but that means more inheritance and more classes.

core-api/src/main/java/com/redhat/lightblue/Response.java
```diff
@@ -62,6 +76,31 @@ public Response(BaseResponse r) {
...
+
+    public boolean isErrorOnResposneSizeTooLarge() {
+        return maxResultSetSizeB > 0;
+    }
+
+    public boolean isWarnOnResponseSizeLarge() {
```
alechenninger: Why one "ResponseSizeLarge" and the other "ResponseSizeTooLarge"?

Also, still typo in "Resposne"

crud/src/main/java/com/redhat/lightblue/hooks/HookManager.java
alechenninger: I don't understand. This method resets the state of this object. Is it tightly coupled to how it expects clients to configure the thresholds? Why not reset to be consistent and decoupled? It's less surprising to the developer. This code is full of enough surprises :-).

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
alechenninger: I see, I read that method as only being used in bulk requests for some reason.

Is it worth coupling whether or not the response is checked to how the hooks work? A performance optimization? Why not set both unconditionally?

core-api/src/main/java/com/redhat/lightblue/Response.java
paterczm: All right, renamed.

crud/src/test/java/com/redhat/lightblue/mediator/MediatorTest.java
alechenninger: Really simple amongst not simple code :-). Simple things break all the time, that's what simple tests are for. Anyway, it's not a huge deal I guess.

crud/src/main/java/com/redhat/lightblue/hooks/HookManager.java
paterczm: Ok, clearing those 2 fields as well.

core-api/src/main/java/com/redhat/lightblue/Response.java
paterczm: Because one is about a warning and the other about error (too large). Fixing typo.

Issue #804, opened 2017-07-27
Remove MediatorHooks


paterczm:
We are not using hooks at the Mediator level. Let's remove them, as this unused logic increases maintenance costs.

PR #805, opened 2017-07-27
Adding MemoryMonitor to enforce memory limits

paterczm:
The limits need to be enforced in multiple locations: Response, BulkExecutionContext, HookManager and IterateAndUpdate (last one from lightblue-mongo). To avoid logic redundancy, creating MemoryMonitor utility.

Note that Mediator level hooks are not used (https://github.com/lightblue-platform/lightblue-core/issues/804).

### DIFF
--- a/util/src/main/java/com/redhat/lightblue/util/MemoryMonitor.java
+++ b/util/src/main/java/com/redhat/lightblue/util/MemoryMonitor.java
@@ -1,0 +1,85 @@
+package com.redhat.lightblue.util;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Utility class for tracking memory usage for given type (assuming it's an immutable value).
+ * You can set thresholds and have it execute callbacks when they are exceeded. See {@link MemoryMonitorTest} for usage.
+ *
+ * @author mpatercz
+ *
+ * @param <T>
+ */
+public class MemoryMonitor<T> {
+
+    @FunctionalInterface
+    public static interface SizeCalculator<T> {
+        public int size(T obj);
+    }
+
+    @FunctionalInterface
+    public static interface ThresholdExceededCallback<T> {
+        public void fire(int currentSizeB, int thresholdB, T obj) throws RuntimeException;
+    }
+
+    public static class ThresholdMonitor<T> {
+
+        final int thresholdB;
+
+        boolean fired = false;
+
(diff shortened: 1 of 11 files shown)


util/src/main/java/com/redhat/lightblue/util/MemoryMonitor.java
```diff
@@ -0,0 +1,85 @@
...
+
+        final ThresholdExceededCallback<T> event;
+
+        public ThresholdMonitor(int thresholdB, ThresholdExceededCallback<T> event) {
+            super();
+            this.thresholdB = thresholdB;
+            this.event = event;
+        }
+
+    }
```
alechenninger: How about  "fire" method that handles calling the callback and changing state? You could also have a method to encapsulate checking its threshold

util/src/main/java/com/redhat/lightblue/util/MemoryMonitor.java
```diff
@@ -0,0 +1,85 @@
...
+        if (m.thresholdB > 0) {
+            this.monitors.add(m);
+        }
+    }
+
+    private void checkThresholdMonitors(T obj) throws RuntimeException {
+        for (ThresholdMonitor<T> m: monitors) {
+            if (dataSizeB > m.thresholdB && !m.fired) {
+                m.fired = true;
+                m.event.fire(dataSizeB, m.thresholdB, obj);
```
alechenninger: May be worth capturing a TODO or something here that the first monitor to throw an exception will stop all other callbacks from firing. As a general purpose class, this is probably surprising, though we know with its current usages it may not be problematic. So, I suggest at least a TODO and/or open an issue to remind ourselves that this should probably catch any thrown exceptions and aggregate them.

util/src/main/java/com/redhat/lightblue/util/MemoryMonitor.java
paterczm: Not sure I understand. You're proposing to add 2 more methods with the logic currently implemented in checkThresholdMonitors?

util/src/main/java/com/redhat/lightblue/util/MemoryMonitor.java
paterczm: Without a use cases this does not justify a TODO/issue imo. I will clarify in javadoc.

PR #808, opened 2017-08-10
Move metrics in lbcore

svkm102:


crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -115,7 +116,13 @@ public Mediator(Metadata md,
      * implementation can perform further validations.
      */
     @StopWatch(loggerName = "stopwatch.com.redhat.lightblue.mediator.Mediator", sizeCalculatorClass = "com.redhat.lightblue.mediator.ResponsePayloadSizeCalculator")
-    public Response insert(InsertionRequest req) {
+    public Response insert(InsertionRequest req, RequestMetrics metrics) {
```
alechenninger: Instead of passing RequestMetrics to each method, can we just pass it in the constructor?

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -662,30 +712,30 @@ protected void wait(BulkExecutionContext ctx) {
...
                 }
             }
         };
     }
 
-    public BulkResponse bulkRequest(BulkRequest requests) {
+    public BulkResponse bulkRequest(BulkRequest requests, RequestMetrics metrics) {
         LOGGER.debug("Bulk request start");
         Error.push("bulk operation");
         ExecutorService executor = Executors.newFixedThreadPool(factory.getBulkParallelExecutions());
```
alechenninger: Might be fine to track the whole bulk request still, without any other information other than it being a bulk request. There isn't really anything else we can add in a sane way I think.

util/src/main/java/com/redhat/lightblue/util/metrics/MetricRegistryFactory.java
```diff
@@ -0,0 +1,53 @@
...
+        METRIC_REGISTRY.register("garbage-collector", new GarbageCollectorMetricSet());
+        METRIC_REGISTRY.register("buffers", new BufferPoolMetricSet(ManagementFactory.getPlatformMBeanServer()));
+        METRIC_REGISTRY.register("memory", new MemoryUsageGaugeSet());
+        METRIC_REGISTRY.register("threads", new ThreadStatesGaugeSet());
+	    
+       final JmxReporter jmxReporter = JmxReporter.forRegistry(METRIC_REGISTRY)
+                .convertRatesTo(TimeUnit.SECONDS)
+                .convertDurationsTo(TimeUnit.MILLISECONDS).build();
+        jmxReporter.start();
+	}
```
alechenninger: Consider lazily initializing this instead of initializing it when the class loads. It will prevent us from starting the threads to report to JMX unnecessarily. `getMetricRegistry` can initialize the singleton and then start the reporter when it does.

util/src/main/java/com/redhat/lightblue/util/metrics/MetricRegistryFactory.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.codahale.metrics.JmxReporter;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.jvm.BufferPoolMetricSet;
+import com.codahale.metrics.jvm.GarbageCollectorMetricSet;
+import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
+import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
+
+/**
+ * Register REST endpoint's and JVM metrics and report them via JMX
```
alechenninger: This does not register REST endpoints.

util/src/main/java/com/redhat/lightblue/util/metrics/RequestMetrics.java
```diff
@@ -0,0 +1,51 @@
...
+    
+    Context startStreamingEntityRequest(String operation, String entity, String version);
+
+    Context startLockRequest(String lockOperation, String domain);
+
+    Context startBulkRequest(String bulkOperation, String entity, String version);
+    
+    void setBulkRequest(boolean bulkRequest);
+    
+    boolean isBulkRequest();
```
alechenninger: Would probably remove the setBulk and isBulk methods (in line with above comments).

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -115,7 +116,13 @@ public Mediator(Metadata md,
      * implementation can perform further validations.
      */
     @StopWatch(loggerName = "stopwatch.com.redhat.lightblue.mediator.Mediator", sizeCalculatorClass = "com.redhat.lightblue.mediator.ResponsePayloadSizeCalculator")
-    public Response insert(InsertionRequest req) {
+    public Response insert(InsertionRequest req, RequestMetrics metrics) {
+        RequestMetrics.Context metricCtx;
+        if (metrics.isBulkRequest()) {
+            metricCtx = metrics.startBulkRequest("insert", req.getEntityVersion().getEntity(), req.getEntityVersion().getVersion());
```
alechenninger: I think if the request is trying to use the default version, this will just have `null` for version. If we initialize this after lightblue has resolved the version, we can get the version that is actually used, which is more useful. I *think* you can get this from below like

```
EntityMetadata md = ctx.getTopLevelEntityMetadata();
md.getVersion().getValue(); // <-- this is the resolved version
```

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -474,7 +511,13 @@ private QueryExpression rewriteUpdateQueryForCompositeSearch(CompositeMetadata m
      * The implementation passes the request to the back-end.
      */
     @StopWatch(loggerName = "stopwatch.com.redhat.lightblue.mediator.Mediator", sizeCalculatorClass = "com.redhat.lightblue.mediator.ResponsePayloadSizeCalculator")
-    public Response find(FindRequest req) {
+    public Response find(FindRequest req, RequestMetrics metrics) {
```
alechenninger: Did you take a stab at also monitoring the findAndStream version? It will have to be implemented a bit differently but it's possible. I can give an example but I have to run right now.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Changed as per suggestion.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -115,7 +116,13 @@ public Mediator(Metadata md,
      * implementation can perform further validations.
      */
     @StopWatch(loggerName = "stopwatch.com.redhat.lightblue.mediator.Mediator", sizeCalculatorClass = "com.redhat.lightblue.mediator.ResponsePayloadSizeCalculator")
-    public Response insert(InsertionRequest req) {
+    public Response insert(InsertionRequest req, RequestMetrics metrics) {
+        RequestMetrics.Context metricCtx;
+        if (metrics.isBulkRequest()) {
+            metricCtx = metrics.startBulkRequest("insert", req.getEntityVersion().getEntity(), req.getEntityVersion().getVersion());
+        } else {
+            metricCtx = metrics.startEntityRequest("insert", req.getEntityVersion().getEntity(), req.getEntityVersion().getVersion());
```
svkm102: So the only reason I kept the measure related code for time being was because I was not sure how exactly the data collected through it is getting used. The mediator code suggests that we are just logging measure data and it's not used anywhere else, but I might be missing something. If it's clear that this is just logging and nothing else we can just remove measure altogether.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Added metrics.startBulkRequest() to track the entire bulk request start and completion.

util/src/main/java/com/redhat/lightblue/util/metrics/MetricRegistryFactory.java
svkm102: Fixed comment.

util/src/main/java/com/redhat/lightblue/util/metrics/MetricRegistryFactory.java
svkm102: Fixed as per suggestion.

util/src/main/java/com/redhat/lightblue/util/metrics/RequestMetrics.java
svkm102: Removed setBulkRequest() and isBulkRequest() as all bulk insert/update/save/delete/find requests will be tracked as normal requests.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Added RequestMetrics as constructor param to mediator().

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Some requests like lock, generate and streaming entity are tracked at rest layer. Lock and generate are not in mediator, so rest was the only layer they can be tracked in.Also since streaming response is handled at rest layer, monitoring streaming requests at that layer is more accurate +  it's easier to differentiate if a request is streaming or not at rest level and monitor is accordingly.

svkm102:
PRs for [lightblue-rest](https://github.com/lightblue-platform/lightblue-rest/pull/300) and [lightblue-audit-hook](https://github.com/lightblue-platform/lightblue-audit-hook/pull/67). This change will also result in updation of some test cases in lightblue-mongo and lightblue-ldap. I'll merge those once this PR is approved.

config/src/main/java/com/redhat/lightblue/config/LightblueFactory.java
```diff
@@ -345,15 +346,15 @@ public Factory getFactory()
         return factory;
     }
 
-    public Mediator getMediator()
+    public Mediator getMediator(RequestMetrics metrics)
```
alechenninger: It doesn't really make sense for this method to accept a RequestMetrics. If you call it again, it's just going to use the first metrics that every called this method. Also, everything else that the Mediator uses is initialized elsewhere.

I would probably accept a RequestMetrics within the LightblueFactory constructor instead.

See also: https://github.com/lightblue-platform/lightblue-rest/pull/300#discussion_r134321943

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -161,12 +164,21 @@ public Response insert(InsertionRequest req) {
...
             }
         } catch (Error e) {
+            if(metricCtx!=null) {
+                metricCtx.markRequestException(e);
+            }
             response.getErrors().add(e);
             response.setStatus(OperationStatus.ERROR);
         } catch (Exception e) {
+            if(metricCtx!=null) {
+                metricCtx.markRequestException(e);
```
alechenninger: Woops. I neglected to realize that pushing down initializing the metricCtx would create opportunities for failure before we initialized it. That's not good. SO, on second thought, how you originally had it was better :-). My bad. The lesson here is that it's important to balance pushing the metrics collection further down for more information, against pulling the metrics further up to get more context. In this case, my suggestion was a smidgen too far down, missing some potential failure scenarios (we want to get as many of those as possible).

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -733,6 +802,7 @@ public BulkResponse bulkRequest(BulkRequest requests) {
             Error.pop();
             return response;
         } finally {
```
alechenninger: We should catch any errors here and add them to the metrics ctx.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -483,7 +529,9 @@ public Response find(FindRequest req) {
         try {
             ctx = newCtx(req, CRUDOperation.FIND);
             ctx.measure.begin("find");
-
+            CompositeMetadata md = ctx.getTopLevelEntityMetadata();
+            metricCtx = metrics.startEntityRequest("find", md.getName(), md.getVersion().getValue());
+            
             StreamingResponse r=_findAndStream(req,ctx);
```
alechenninger: We can handle streaming in the mediator, not HTTP (lightblue-rest) layer.

We want to hook into when the document stream is fully consumed or closed. Closed might be the best indicator.

One simple way I can think of to do this is to write a MeteredDocumentStream that delegates to another stream entirely with the addition that when the close() method is called, the MeteredDocumentStream __also__ ends a metric ctx.

Then in `_findAndStream`, instead of:

```
            CRUDFindResponse result = finder.find(ctx, req.getCRUDFindRequest());
            ctx.measure.end("finder.find");
            
            if(!ctx.hasErrors()) {
                ctx.setStatus(OperationStatus.COMPLETE);
                response.documentStream=ctx.getDocumentStream();
                if(ctx.isComputeCounts()) {
                    response.matchCount=result.getSize();
                }
            } else {
                ctx.setStatus(OperationStatus.ERROR);
            }
```

...you'd do something like (accepting a `RequestMetric.Context metricCtx` as a method argument):

```
            CRUDFindResponse result = finder.find(ctx, req.getCRUDFindRequest());
            ctx.measure.end("finder.find");
            
            if(!ctx.hasErrors()) {
                ctx.setStatus(OperationStatus.COMPLETE);
                response.documentStream=new MeteredDocumentStream(ctx.getDocumentStream(), metricCtx);
                if(ctx.isComputeCounts()) {
                    response.matchCount=result.getSize();
                }
            } else {
                ctx.setStatus(OperationStatus.ERROR);
                // Do we need to do this in more places?
                ctx.getErrors().forEach(metricCtx::markRequestException);
                metricCtx.endRequestMonitoring();
            }
```

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -161,12 +164,21 @@ public Response insert(InsertionRequest req) {
...
             response.setStatus(OperationStatus.ERROR);
         } catch (Exception e) {
+            if(metricCtx!=null) {
+                metricCtx.markRequestException(e);
+            }
             response.getErrors().add(Error.get(CrudConstants.ERR_CRUD, e));
             response.setStatus(OperationStatus.ERROR);
         } finally {
+            if(metricCtx!=null) {
+                metricCtx.endRequestMonitoring();
```
alechenninger: Some exceptions might not get metered if they don't bubble up to the catch blocks here. So maybe what we want to do is, instead of adding to each catch block, rely on the errors being added to the response as they should be. Then, in the finally block, go through each of the errors in the Response and mark them with the request ctx. This ensures we are tracking response errors thoroughly.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
alechenninger: Should probably be adding these to the response I guess too? Now I'm wondering why that wasn't being done already.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
alechenninger: Related to getting more context, there is another option we should consider: accepting a RequestMetrics.Context in the Mediator methods like find, insert, etc.

This could get us the best of both worlds: broad coverage of the lifecycle of a request, but also coverage of the innards (like within bulk requests), at the cost of some complexity.

However, I think it might be worth it. For example, right now, if a request fails to parse, we won't get any metrics on that, because it never makes it to the Mediator. However, if we started tracking before we got to the Mediator, we could capture that.

Within bulk requests, then, it would do both: use the provided context for the bulk request, but also create new contexts to track the individual requests within it.

What do you think?

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -98,13 +99,13 @@
 
     public final Metadata metadata;
     public final Factory factory;
-
-    public Mediator(Metadata md,
-                    Factory factory) {
+    public final RequestMetrics metrics;
+   
+    public Mediator(Metadata md, Factory factory, RequestMetrics metrics) {
```
bserdar: This is not good. Why is metrics being passed in ctor? Ideally, metrics should be in the Factory, or LightblueConfiguration, and not passed like this. 

Does this metrics class work only at the mediator level? If so, a cleaner approach would be to remove all metrics related code from the mediator, and write a MediatorWithMetrics class extending mediator, and add metrics code there before and after the original mediator methods. You can return MediatorWithMetrics from LightblueFactory based on config.

config/src/main/java/com/redhat/lightblue/config/LightblueFactory.java
svkm102: Altered this approach to pass metricCtx to all mediator commands, for better metric tracking.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: I too feel this should give more context information in each metric and will handle errors metrics more efficiently. Modified this PR accordingly.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Fixed as per suggestions.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Removed metrics as constructor param.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: I guess since all individual bulk requests (find/update/delete/save) are tracking their own errors, there is no tracking present at the overall bulkrequest level. Also BulkResponse object does not have ability to add errors, so this was left as is. Tracking errors at this level will require changing BulkResponse composition.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -115,7 +114,7 @@ public Mediator(Metadata md,
      * implementation can perform further validations.
      */
     @StopWatch(loggerName = "stopwatch.com.redhat.lightblue.mediator.Mediator", sizeCalculatorClass = "com.redhat.lightblue.mediator.ResponsePayloadSizeCalculator")
-    public Response insert(InsertionRequest req) {
+    public Response insert(InsertionRequest req, RequestMetrics.Context metricCtx) {
```
bserdar: I don't like metrics context being passed like this. I can think of a few options that might work better.

One option is to use OperationContext to keep RequestMetrics.Context. But, operation context is constructed in mediator calls (newCtx), so extending mediator with a new newCtx method that sets the metrics could work.

Another method is to use a ThreadLocal in RequetMetrics.Context to set it in rest layer, and get it in newCtx() in mediator.

The code needs to handle null request metrics. It can be done by using a noop metrics implementation.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
alechenninger: I see. I guess the REST layer takes care of any errors there, so we can track that there.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -701,28 +721,32 @@ public BulkResponse bulkRequest(BulkRequest requests) {
...
                     // ordered - only consecutive finds in parallel
                     if (req.getOperation() == CRUDOperation.FIND) {
-                        ctx.futures[i] = executor.submit(getFutureRequest(req));
+                        ctx.futures[i] = executor.submit(getFutureRequest(req, metrics));
                     } else {
                         wait(ctx);
                         switch (req.getOperation()) {
                             case INSERT:
-                                ctx.setResponseAt(i, insert((InsertionRequest) req));
+                            	requestCtx = metrics.startEntityRequest("insert", req.getEntityVersion().getEntity(), req.getEntityVersion().getVersion());
```
alechenninger: So per suggestion above, I would probably start and stop tracking here. This way we don't have to pass around the metricCtx everywhere, and we could do the same thing in REST layer, consistent with other requests.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -483,8 +490,9 @@ public Response find(FindRequest req) {
         try {
             ctx = newCtx(req, CRUDOperation.FIND);
             ctx.measure.begin("find");
-
-            StreamingResponse r=_findAndStream(req,ctx);
+            CompositeMetadata md = ctx.getTopLevelEntityMetadata();
```
alechenninger: This looks like it's unused now?

alechenninger:
> then duplicate the work the REST layer does because this time it's calling the mediator, not the REST layer

One thought here: we could maybe add a method to the RequestMetrics.Context interface that ends the tracking with a Response object. Then we can have this logic reusable in both places.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: So we decided to track requests at REST layer and parse mediator exceptions via response objects itself. This approach will remove passing or setting of this metricCtx object altogether.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -172,6 +171,8 @@ public Response insert(InsertionRequest req) {
                 METRICS.debug("insert: {}",ctx.measure);
             }
             Error.pop();
+            markRequestErrors(response.getErrors(), metricCtx);
+            metricCtx.endRequestMonitoring();
```
svkm102: Fixed as per suggestion. Mediator will only track the individual bulkrequests around their invocations.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Ya, forgot to delete that. Thanks for catching it :)

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Fixed as per suggestion.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -858,4 +883,10 @@ private boolean checkQueryAccess(OperationContext ctx, QueryExpression query) {
         }
         return dataErrors;
     }
+    
+	private void markRequestErrors(List<Error> errors, RequestMetrics.Context metricCtx) {
```
svkm102: Added this to Context.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -662,30 +663,47 @@ protected void wait(BulkExecutionContext ctx) {
...
+                        metricCtx = metrics.startEntityRequest("update", req.getEntityVersion().getEntity(), req.getEntityVersion().getVersion());
+                        resp = update((UpdateRequest) req);
+                        metricCtx.markAllErrorsAndEndRequestMonitoring(resp.getErrors());
+                        return resp;
                     case SAVE:
-                        return save((SaveRequest) req);
+                        metricCtx = metrics.startEntityRequest("save", req.getEntityVersion().getEntity(), req.getEntityVersion().getVersion());
+                        resp = save((SaveRequest) req);
+                        metricCtx.markAllErrorsAndEndRequestMonitoring(resp.getErrors());
+                        return resp;
```
alechenninger: This looks good. For what it's worth, you could probably refactor so that we store the response in the switch, then outside of the switch we have a single `metricCtx.markAllErrorsAndEndRequestMonitoring(resp.getErrors());` and a single return. It reduces duplication a little bit. Not a huge deal though.

Also, we could consider adding an overload to `startEntityRequest` that accepts a `CRUDOperation`. Then we could have a single `startEntityRequest` call as well.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -695,34 +713,47 @@ public BulkResponse bulkRequest(BulkRequest requests) {
...
+                                metricCtx = metrics.startEntityRequest("update", req.getEntityVersion().getEntity(), req.getEntityVersion().getVersion());
+                                resp = update((UpdateRequest) req);
+                                metricCtx.markAllErrorsAndEndRequestMonitoring(resp.getErrors());
+                                ctx.setResponseAt(i, resp);
                                 break;
                             case SAVE:
-                                ctx.setResponseAt(i, save((SaveRequest) req));
+                                metricCtx = metrics.startEntityRequest("save", req.getEntityVersion().getEntity(), req.getEntityVersion().getVersion());
+                                resp = save((SaveRequest) req);
+                                metricCtx.markAllErrorsAndEndRequestMonitoring(resp.getErrors());
```
alechenninger: Similar to future code, this looks good. We could do similar refactoring to reduce duplication would be my only suggestion.

svkm102: Refractored as suggested.

crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
svkm102: Refractored as suggested.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        @Override
+        public void endRequestMonitoring() {
+            if (!ended) {
+                ended = true;
+                activeRequests.dec();
+                context.stop();
+            } else {
+                LOGGER.warn("Request already ended");
```
alechenninger: Would be a good idea to add the metric nameSpace to this warn so that we know what metric was "double ended."

util/src/main/java/com/redhat/lightblue/util/metrics/RequestMetrics.java
```diff
@@ -0,0 +1,52 @@
...
+import com.redhat.lightblue.util.Error;
+
+public interface RequestMetrics {
+    /**
+     * Start timers and counters for the request. Use the returned context to
+     * complete the request and optionally mark errors if they occur.
+     *
+     * <p>
+     * The returned context itself is not completely thread safe, it is expected to
+     * be used by one and only one thread concurrently.
```
alechenninger: Might want to put this note about thread safety on the Context interface javadoc and/or doc other APIs a bit more.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
svkm102: Done.

util/src/main/java/com/redhat/lightblue/util/metrics/RequestMetrics.java
svkm102: Added javadocs for each operation.

svkm102:
Added test case for metrics in BulkTest

alechenninger:
Thanks for adding that test! There is a request I have for rest that will require an update here. It'd also be nice to see some more test coverage here.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
```diff
@@ -0,0 +1,145 @@
...
+    
+    @Override
+    public Context startLockRequest(String lockOperation, String domain) {
+        return new DropwizardContext(name(API, "lock", domain, lockOperation));
+    }
+
+    @Override
+    public Context startSavedSearchRequest(String searchName, String entity, String version) {
+        return new DropwizardContext(name(API, "savedsearch", searchName, entity, version));
+    }
```
alechenninger: For all of these that use version, if version is null, this will roll up to to just "operation.entity" which means technically we lose the ability to distinguish between default version searches and other versions. How about if version is null, using the string "default" instead?

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
```diff
@@ -0,0 +1,145 @@
...
+
+        @Override
+        public void markRequestException(Error e) {
+            String errorCode = e.getErrorCode().replaceAll(REGEX, ".");
+            metricsRegistry.meter(errorNamespace(metricNamespace, e, errorCode)).mark();            
+        }
+        
+        @Override
+        public void markRequestException(Exception e) {
+        	String errorMessage = e.getMessage().replaceAll(REGEX, ".");
```
alechenninger: Using the error message in the metric could get really messy. These are arbitrary strings, and could be very long. For Error, we're using the error code–a //very// different thing, since these are well known, and short.

util/src/test/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetricsTest.java
```diff
@@ -0,0 +1,115 @@
...
+        context.markRequestException(new Exception("bulk request exception"));
+        Meter exceptionMeter = metricsRegistry.meter("api.bulk.requests.exception.Exception.bulk.request.exception");
+        Assert.assertEquals(1, exceptionMeter.getCount());
+    }
+    
+    @Test
+    public void testMarkRequestExceptionWithRegex() {
+        DropwizardRequestMetrics.Context context = requestMetrics.startLockRequest("lockcommand", "domain");
+        context.markRequestException(new NullPointerException("this@tests$the_regex#pattern"));
+        Meter exceptionMeter = metricsRegistry.meter("api.lock.domain.lockcommand.requests.exception.NullPointerException.this.tests.the.regex.pattern");
```
alechenninger: Like I mentioned above, it could get really messy to use the entire error message in the metric, so this test is invalid for that reason. However, I would still test the regex handling just with Error's error codes instead.

alechenninger:
Found two things doing a final once over, otherwise looks good. Just fix those (make sure any changes are tested of course) and we're good to go.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
svkm102: Added a null check to replace null version with default as suggested.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
svkm102: Removed error message from metric namespace in case of Exception.

util/src/test/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetricsTest.java
svkm102: Updated test cases and included some new tests for better coverage.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
alechenninger: Thanks!

We found one more thing that we need to address with version. Basically, right now the convention is that versions use semver like "1.0.0". However "." is a meaningful character with graphite metrics, so this creates 3 path segments "1", "0", and "0" where the default version will only have one (just "default"). When you do queries with graphite, you use wild cards per path segment, per the docs:

> All wildcards apply only within a single path element. In other words, they do not include or cross dots (.). Therefore, servers.* will not match servers.ix02ehssvc04v.cpu.total.user, while servers.*.*.*.* will.

So that means it'd be pretty difficult if not impossible to have a query look at all versions including default.

The solution I believe is rather simple: let's just replace all of the "." with some other character like "_" in version. Then both default and explicit versions will use only one path segment.

svkm102: Replaced "." with "_" in version. Also removed the regex pattern used earlier to format error code with a "." separated value and replaced it with a replace(":","-") . So now the error "mongo-crud:DatabaseError" will be reported as "mongo-crud-DatabaseError" (a single entity for easier graphite query), instead of mongo.crud.DatabaseError.

The only reason I included regex in first place was because having some special characters in metric namespace caused it to display a bit differently in visualvm. For eg:- 
without special characters - api.find.user.1.2.3.requests.active 
with special characters - "api.find.user.1.2.3.requests.exception.Error.mongo-metadata:UnknownVersion" (notice the opening and closing quote)

While doing the version changes, found that the quotes get added to namespace only for a few special characters and not for all.So if we have "_"  or " - " in namespace, it displays normaly  and only ":" was causing issues. 
Since ":" is the only other special character we use in our error codes apart from "-", replacing just this one shou;ld suffice for our needs.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
```diff
@@ -72,20 +69,25 @@ private static String errorNamespace(String metricNamespace, Throwable exception
         return e.getClass();
     }
 
-    private String checkNullVersion(String version) {
+    /**
+     * If version is null,replace with default.
+     * Also replace all . in version with _ as graphite treats . as standard separator and can cause issues in queries.
+     * 
+     */
+    private String checkVersion(String version) {
```
alechenninger: How about `cleanVersion` or `formatVersion`? `checkVersion` implies it is doing some kind of validation and would expect it to throw an exception

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
```diff
@@ -131,7 +133,9 @@ public void endRequestMonitoring() {
 
         @Override
         public void markRequestException(Error e) {
-            String errorCode = e.getErrorCode().replaceAll(REGEX, ".");
+            // Presence of : in metricnamespace makes it's display a bit weird in visualvm.
+            // This might not effect the metric in graphite, but replacing it as a precaution.
+            String errorCode = e.getErrorCode().replace(":", "-");
```
alechenninger: How about replacing this with _ like we do with version? It might be surprising to use a different replacement character I think.

We also still might want to replace a few other things, like any whitespace that might creep into error codes. AFAIK there isn't a well-defined set of valid characters in error codes, so I did like that you were being relatively cautious before.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
svkm102: Fixed name with formatVersion.

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
svkm102: Added Regex pattern matching back to error code. This time it replaces any special character with "_"

PR #809, opened 2017-08-18
Health check enhancements

derek63:


util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public Context startHealthRequest() {
+        return new DropwizardContext(name(API, "checkHealth"));
+    }
+
+    @Override
+    public Context startDiagnosticsRequest() {
+        return new DropwizardContext(name(API, "checkDiagnostics"));
+    }
+
```
svkm102: Not sure if collecting metrics on /health and /diagnostics will be of much use. I don't think we would care about the timing of these requests or using these metrics to configure any alerts on these alert endpoints themselves.
There's no harm in keeping them (for consistency), but if they are not going to be used, it's better to keep these endpoints un monitored.

derek63: They were initially included for consistency's sake, and even though I am not exactly sure how these will be used yet, I think we should leave them in there.   I can't think of any uses for the diagnostics one at the moment, however perhaps the health one can serve as a baseline /control for other monitored find requests, as it does a find on an entity with minimal data and few clients.  I guess if we find they're not useful at all we can just remove these later.

derek63:
> Would you be willing to also tackle adding some more information to CRUDHealth?

Sure, I updated CRUDHealth, created https://github.com/lightblue-platform/lightblue-mongo/pull/380, and updated https://github.com/lightblue-platform/lightblue-rest/pull/301

util/src/main/java/com/redhat/lightblue/util/metrics/DropwizardRequestMetrics.java
alechenninger: The latencies won't be useful really since the health checks do the work in the background, the latencies aren't measuring any remote calls, only in-memory work which will take nanoseconds or less. But the counts are still useful I think, so we can see how often clients are asking about health or diagnostics, potentially spotting if load balancers are misconfigured or clients are asking way too often.

PR #814, opened 2017-10-05
Setting status on streaming responses properly

paterczm:
Also some refactoring to ensure status is always set on response.

### DIFF
--- a/crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
+++ b/crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
@@ -119,7 +119,7 @@ public class Mediator {
     public Response insert(InsertionRequest req) {
         LOGGER.debug("insert {}", req.getEntityVersion());
         Error.push("insert(" + req.getEntityVersion().toString() + ")");
-        Response response = new Response(factory.getNodeFactory());
+        Response response = new Response(factory.getNodeFactory(), OperationStatus.ERROR);
         OperationContext ctx=null;
         try {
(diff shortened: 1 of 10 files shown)


crud/src/main/java/com/redhat/lightblue/mediator/Mediator.java
```diff
@@ -119,7 +119,7 @@ public Mediator(Metadata md,
     public Response insert(InsertionRequest req) {
         LOGGER.debug("insert {}", req.getEntityVersion());
         Error.push("insert(" + req.getEntityVersion().toString() + ")");
-        Response response = new Response(factory.getNodeFactory());
+        Response response = new Response(factory.getNodeFactory(), OperationStatus.ERROR);
```
derek63: I see where these statuses are being set to COMPLETE later if there are no errors, any particular reason we went with "initialize as ERROR, override with COMPLETE" instead of "initialize COMPLETE and override with ERROR" or introduce a third status to indicate "response not yet submitted/received" , maybe PENDING or something.  I guess it just seems strange to assume failure, but maybe there is a reason I am missing (or will see in one of the related PRs)

paterczm: I don't know any other reason than that being a convention already, e.g. https://github.com/lightblue-platform/lightblue-core/pull/814/files/ee4e61f124575d78e2da4aa7b3474c49607b76af#diff-3dad0be22f493d3332539c3b96790af0L483.

PR #820, opened 2018-04-20
A flexible way to initialize ControllerFactory

paterczm:
This change is to allow parametrization of ControllerFactory. Not sure if it has to be initialized lazily, but keeping existing behavior.

The use case I have in mind is passing MongoClient to Lightblue, as opposed to letting Lightblue initialize it itself. Helpful when MongoClient is managed by other means (e.g. spring).


paterczm:
Closing, there will be a slightly different solution.

PR #821, opened 2018-07-31
 Exclude negative ids from ArrayElementIdChecker

svkm102:
This should allow us to handle cases where we deliberately add negative ids to handle some specific scenarios and those ids can repeat in the document without any side effects.


crud/src/main/java/com/redhat/lightblue/crud/validator/ArrayElementIdChecker.java
```diff
@@ -64,7 +64,7 @@ public void checkConstraint(ConstraintValidator validator,
             iterator.next();
             Path fieldName=iterator.getCurrentKey();
             JsonNode value=iterator.getCurrentValue();
-            if(!values.add(t.fromJson(value)))
+            if(value.asInt(0) >= 0 && !values.add(t.fromJson(value)))
```
bserdar: Array element id is in general, not an int. Why is this necessary?

alechenninger:
IIRC array element id uniqueness is used for other algorithms in lightblue, I don't think we can just remove that constraint without other consequences, though @bserdar would know better than I would.

Can you generate another UUID for the element id instead?

bserdar:
You can initialize the element id field with a uuid using an initializer that won't overwrite.

PR #823, opened 2018-10-09
Monitor memory usage of composite finds

alechenninger:
- [x] Add MemoryMonitor to ExecutionContext (used in composite finds)
- [x] Monitor memory from composite find pipeline steps that aggregate results in memory
- [x] Allow controlling composite find result limits independently from other reads


paterczm:
Looks good.

I would set default maxExecutionContextSizeForCompositeFindB higher than 50 MB to start with, because it's a potentially breaking change. The ExecutionContextIsLarge warning should give you an idea of what values to expect and tweak the limit accordingly. No need to change the default in the PR, just saying.

alechenninger:
Good point, I agree. Thank you!

alechenninger:
Turns out I had set it to 100mb but forgot to update the comment. Doh. I'll go with that default, but as we roll it out I may use a yet higher value to be safe. We'll see. Thanks again.

